PR #8, opened 2026-04-15
fix: replace prompt-type SessionStart/PostCompact hooks with command-type

### LINKED ISSUE #7: SessionStart:resume and PostCompact prompt-type hooks fail with "ToolUseContext is required for prompt hooks"
## Summary

The `SessionStart:resume` and `PostCompact` hooks in `hooks/hooks.json` use `type: "prompt"`, which errors out on Claude Code CLI v2.1.108 because `ToolUseContext` is not provided in those contexts.

## Reproduction

1. Install claude-obsidian v1.4.3 via plugin marketplace
2. Set up a vault (CLAUDE.md + wiki/hot.md)
3. Start a new Claude Code session, then resume it (e.g., after `/clear` or reopening)
4. On every resume, Claude Code shows:

```
SessionStart:resume hook error
Failed to run: ToolUseContext is required for prompt hooks. This is a bug.
```

## Root Cause

Current `hooks.json`:

```json
"SessionStart": [{
  "matcher": "startup|resume",
  "hooks": [
    { "type": "command", "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true" },
    { "type": "prompt",  "prompt": "If a vault is configured for this session..." }
  ]
}],
"PostCompact": [{
  "hooks": [{ "type": "prompt", "prompt": "Hook-injected context does not survive..." }]
}]
```

Prompt-type hooks require a `ToolUseContext` runtime object. Claude Code does not supply one for `SessionStart:resume` or `PostCompact`, so they fail with the error above.

This is the same class of bug that was fixed for the `Stop` hook in commit `e54b419` ("fix: replace prompt-type Stop hook with command-type to prevent infinite loop"). The fix was not applied to `SessionStart` or `PostCompact`.

## Impact

Functionally minor — the preceding `command` hook in `SessionStart` already executes `cat wiki/hot.md`, so hot cache is still injected via stdout. The prompt hook was a defensive redundancy that now just produces an error banner on every resume.

However, `PostCompact` has **only** a prompt hook, so the post-compaction re-injection of hot cache silently breaks on long sessions that trigger auto-compact.

## Suggested Fix

Either:

1. **Drop the redundant prompt hooks.** `SessionStart`'s `command` hook already does the cat; the prompt version is defense-in-depth that's currently broken anyway. For `PostCompact`, replace with a `command` hook that re-cats `wiki/hot.md`.

2. **Wait for upstream Claude Code to supply `ToolUseContext` on resume/postcompact.** But this depends on Anthropic's timeline.

Option 1 is the pragmatic fix and matches what was already done for `Stop` in e54b419.

## Environment

- Claude Code CLI: v2.1.108
- claude-obsidian plugin: v1.4.3 (latest release)
- OS: macOS (darwin 25.3.0)

crowscc:
## Summary

Fixes the `SessionStart:resume hook error: Failed to run: ToolUseContext is required for prompt hooks. This is a bug.` error reported in #7.

## Root Cause

`SessionStart` (second hook) and `PostCompact` use `type: "prompt"`, which requires a `ToolUseContext` runtime object that is not available during session resume or post-compaction. Same class of issue as the one fixed for the `Stop` hook in e54b419 — prompt-type hooks depend on LLM-conversation runtime state that isn't always available, while command-type hooks only depend on the shell.

## Changes

- **SessionStart**: drop the redundant second (prompt) hook. The first command hook already runs `cat wiki/hot.md`, which injects hot cache content via stdout. The prompt hook was defense-in-depth that never worked on resume paths.
- **PostCompact**: convert prompt hook to command hook with the same `cat wiki/hot.md` pattern. Preserves the "re-inject hot cache after compaction" functionality using the same technique already validated by the Stop fix.

## Diff

```diff
 "SessionStart": [{
   "matcher": "startup|resume",
   "hooks": [
     { "type": "command", "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true" }
-    ,{ "type": "prompt", "prompt": "If a vault is configured..." }
   ]
 }],
 "PostCompact": [{
   "matcher": "",
   "hooks": [
-    { "type": "prompt", "prompt": "Hook-injected context does not survive..." }
+    { "type": "command", "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true" }
   ]
 }]
```

## Verification

- [x] JSON validated with `python3 -m json.tool hooks/hooks.json`
- [x] No remaining `"type": "prompt"` entries in hooks.json
- [x] Fix pattern matches the existing `Stop` hook fix (e54b419) — same approach: replace prompt-type with command-type that writes prompt-like output to stdout when needed
- [x] Untouched: `PostToolUse` auto-commit, `Stop` WIKI_CHANGED nudge
- [ ] End-to-end runtime verification (starting + resuming a Claude Code session with patched plugin) — **not performed by author**; intentionally left for maintainer/reviewer. The change is a small config transform and relies on the same command-stdout pattern already validated in e54b419.

## Behavioral Notes

- **Hot cache injection**: unchanged — still works on every session start (command hook cats `wiki/hot.md`).
- **Post-compaction restore**: now functional (was silently broken before due to the `ToolUseContext` error on the prompt hook).
- **Lost nuance**: the original prompt hooks included "silently, do not announce" language directed at the LLM. With command-type hooks, this instruction is no longer emitted. In practice, command hook stdout is injected as context, and the vault's `CLAUDE.md` governs how Claude describes it. No observable difference expected in a typical Obsidian vault session.

## Rollback

Revertible by `git revert <sha>`. No state migration, no downstream dependencies. The prior behavior (with error banner) would be restored.

## Closes

Fixes #7

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -7,10 +7,6 @@
           {
             "type": "command",
             "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true"
-          },
-          {
-            "type": "prompt",
-            "prompt": "If a vault is configured for this session (check CLAUDE.md for VAULT_PATH or a wiki/ folder in the current directory), silently read wiki/hot.md to restore recent context. If wiki/hot.md does not exist, do nothing. This is a non-vault session. Do not announce this. Do not report what you read. Just have the context available."
           }
         ]
       }
@@ -20,8 +16,8 @@
         "matcher": "",
         "hooks": [
           {
-            "type": "prompt",
-            "prompt": "Hook-injected context does not survive context compaction. If wiki/hot.md exists in the current directory, silently re-read it now to restore the hot cache. Do not announce this."
+            "type": "command",
+            "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true"
           }
         ]
       }


crowscc:
Closing this PR — our original approach (replacing `prompt` hooks with plain `cat` stdout) has a fundamental flaw.

**What we found after further investigation:**

1. **`cat` stdout doesn't actually inject into model context** on `SessionStart` — despite [official docs](https://code.claude.com/docs/en/hooks) claiming it does for `SessionStart` and `UserPromptSubmit`. We verified locally: Claude had to manually `Read` files instead of answering from context. This appears to be a regression or doc/implementation mismatch in v2.1.110.

2. **`PostCompact` cannot inject context via any mechanism** — official docs explicitly state "`additionalContext` ignored; cannot reach model" for `PostCompact`. So neither our `cat` approach nor @FreddieMcHeart's JSON `additionalContext` approach would work for PostCompact.

3. **The `prompt` hook type is broken on both `startup` AND `resume`** — not just resume as we originally reported in #7. The `ToolUseContext` error fires on fresh sessions too (confirmed on v2.1.110).

**Bottom line:** The correct fix needs to account for these Claude Code platform limitations. A new PR should:
- Use JSON `additionalContext` for `SessionStart` (the only mechanism that might work)
- Accept that `PostCompact` hot cache re-injection is currently impossible at the platform level
- Also fix `Stop` hook (current `echo` stdout goes to debug log only per docs, not to model)

See the full analysis in #7. Leaving the issue open for a better fix.

Issue #10, opened 2026-04-16
Bug: `prompt` type hook in SessionStart causes "ToolUseContext is required" error


jhsong-musinsa:
## Description

The `SessionStart` hook in `hooks/hooks.json` includes a `prompt` type hook, which causes the following error on every session start:

```
SessionStart:startup hook error
Failed to run: ToolUseContext is required for prompt hook
This is a bug.
```

## Root Cause

In `hooks/hooks.json`, the `SessionStart` event has two hooks:

```json
"SessionStart": [
  {
    "matcher": "startup|resume",
    "hooks": [
      {
        "type": "command",
        "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true"
      },
      {
        "type": "prompt",
        "prompt": "If a vault is configured for this session..."
      }
    ]
  }
]
```

The `prompt` type hook requires a `ToolUseContext`, which is not available during the `SessionStart` event. Claude Code only provides `ToolUseContext` for tool-use-scoped events (e.g., `PostToolUse`), not for session lifecycle events.

## Impact

- Error message displayed on every session start
- The `command` hook (which reads `wiki/hot.md`) still runs successfully, so functionality is not lost — the error is cosmetic but noisy

## Suggested Fix

Remove the `prompt` hook from the `SessionStart` event. The `command` hook already outputs the contents of `wiki/hot.md`, which achieves the same goal of restoring context at session start.

```diff
 "SessionStart": [
   {
     "matcher": "startup|resume",
     "hooks": [
       {
         "type": "command",
         "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true"
-      },
-      {
-        "type": "prompt",
-        "prompt": "If a vault is configured for this session (check CLAUDE.md for VAULT_PATH or a wiki/ folder in the current directory), silently read wiki/hot.md to restore recent context. If wiki/hot.md does not exist, do nothing. This is a non-vault session. Do not announce this. Do not report what you read. Just have the context available."
       }
     ]
   }
 ]
```

The same issue also exists in the `PostCompact` event, which has a `prompt` hook that may encounter the same problem depending on the Claude Code version.

## Environment

- Claude Code CLI
- Plugin version: 1.4.3
- OS: macOS

jhsong-musinsa:
Closing as duplicate of #7 (reported two days earlier with the same symptom and root cause).

[...]

PR #13, opened 2026-04-17
fix: remove prompt-type hooks from SessionStart and PostCompact

### LINKED ISSUE #7: SessionStart:resume and PostCompact prompt-type hooks fail with "ToolUseContext is required for prompt hooks"

jhsong-musinsa:
## Summary

Closes #7 (also reported as duplicate #10) — removes the `Failed to run: ToolUseContext is required for prompt hook` banner on session start by emitting the documented `hookSpecificOutput.additionalContext` JSON from a command-type `SessionStart` hook, and unregisters `PostCompact` because it has no supported injection mechanism.

> This PR supersedes the approach proposed in [#8](https://github.com/AgriciDaniel/claude-obsidian/pull/8), which its author closed after finding that plain `cat` stdout is not actually injected into the model context on `SessionStart`.

## Changes

| File | Change |
|---|---|
| `hooks/hooks.json` | `SessionStart` emits `{hookSpecificOutput:{hookEventName,additionalContext}}` via `jq`, resolving `${CLAUDE_PROJECT_DIR:-$PWD}/wiki/hot.md`; guarded with `[ -f "$HOT" ]`, `command -v jq`, and `\|\| true`. `PostCompact` event fully removed. |
| `hooks/README.md` | Events table updated; new "Why no `PostCompact` / `prompt`-type hooks" and "Dependencies" sections; stale plain-`cat` workaround removed. |

## Why this approach

Per the [official hooks reference](https://code.claude.com/docs/en/hooks):

- `SessionStart` supports context injection **only** through `hookSpecificOutput.additionalContext` — plain stdout is not guaranteed to reach the model.
- `PostCompact` is listed under events with **"None" decision control** — no `additionalContext`, no prompt injection, no way to reach the model. Keeping any hook there can only produce side effects or errors.
- `prompt`-type hooks depend on a tool-use-scoped runtime (`ToolUseContext`) that session lifecycle events do not supply, which is why the banner appeared on every session start.

## Verified

- [x] `hooks/hooks.json` parses as valid JSON
- [x] Hook command exits 0 in a populated vault (`wiki/hot.md` present) and produces the documented JSON payload
- [x] Hook command exits 0 in a non-vault directory (`wiki/hot.md` absent)
- [x] Hook command exits 0 when `jq` is not on `PATH` (feature degrades silently)
- [x] JSON escaping correct for content containing `"` and `\` (spot-checked via `jq --rawfile`)
- [x] No remaining `"type": "prompt"` entries in `hooks.json`; other hooks (`PostToolUse`, `Stop`) unchanged

## Unverified (requires maintainer env)

- [ ] End-to-end: install the plugin in a fresh Claude Code session and confirm (a) the `ToolUseContext` banner is gone and (b) `wiki/hot.md` content is available in context (e.g., ask Claude for the `Last Updated` value without it running a Read tool)
- [ ] Interaction with the upstream plugin-hook STDOUT bug (`anthropics/claude-code#10875`), still documented in README as a fallback pointer

## Rollback

`git revert <sha>` is safe — no state migration, no downstream dependencies. The prior behavior (error banner on startup, no-op on compact) would be restored.

### DIFF
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -6,21 +6,24 @@ Plugin hooks for the claude-obsidian wiki vault. All hooks are defined in `hooks
 
 | Event | Type | Purpose |
 |---|---|---|
-| `SessionStart` | command | Loads `wiki/hot.md` into context via `[ -f wiki/hot.md ] && cat wiki/hot.md` (works for non-vault sessions without erroring). Matcher: `startup\|resume`. |
-| `PostCompact` | command | Re-loads `wiki/hot.md` via `cat` after context compaction. Hook-injected context does NOT survive compaction (only `CLAUDE.md` does), so this hook restores the hot cache mid-session. |
-| `PostToolUse` | command | Auto-commits any wiki/ or .raw/ changes after Write or Edit tool calls. Guarded by `[ -d .git ]` so it never errors in non-git directories, and by `git diff --cached --quiet` so it never creates empty commits. |
-| `Stop` | prompt | Updates `wiki/hot.md` at the end of every Claude response with a brief summary of what changed. |
+| `SessionStart` | command | Injects `wiki/hot.md` into the model context by emitting the documented `hookSpecificOutput.additionalContext` JSON schema. Resolves the vault path via `${CLAUDE_PROJECT_DIR:-$PWD}/wiki/hot.md`, exits silently when the file is absent (non-vault sessions) or `jq` is not installed. Matcher: `startup\|resume`. |
+| `PostToolUse` | command | Auto-commits any `wiki/` or `.raw/` changes after Write or Edit tool calls. Guarded by `[ -d .git ]` so it never errors in non-git directories, and by `git diff --cached --quiet` so it never creates empty commits. |
+| `Stop` | command | Emits a `WIKI_CHANGED:` nudge when files under `wiki/` have changed this session, prompting Claude to refresh `wiki/hot.md`. Guarded by `[ -d wiki ]`, `[ -d .git ]`, and `grep -q '^wiki/'` so it stays silent otherwise (and avoids the infinite-loop regression caused by the earlier prompt-type version; see commit `e54b419`). |
 
-## Known Issue: Plugin Hooks STDOUT Bug
+## Why no `PostCompact` / `prompt`-type hooks
 
-`anthropics/claude-code#10875` documents that **plugin hook STDOUT may not be captured** by Claude Code, while identical inline hooks in `settings.json` work correctly.
+Claude Code's official [hooks reference](https://code.claude.com/docs/en/hooks) lists `PostCompact` under events with **"None" decision control** — it has no mechanism (no `additionalContext`, no prompt injection) to reach the model. Registering any hook there can only produce side effects (or errors). The earlier `prompt`-type hook on `PostCompact`/`SessionStart` failed with `Failed to run: ToolUseContext is required for prompt hook` because prompt hooks depend on a tool-use-scoped runtime that session lifecycle events do not provide. We therefore unregister `PostCompact` entirely and rely on the supported `SessionStart` JSON schema.
 
-**Impact**: If this bug is active in your Claude Code version, the command-type SessionStart and PostCompact hooks may not inject `wiki/hot.md` into context as expected.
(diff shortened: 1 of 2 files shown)


autorunman22:
after updating `hooks.json` i still get the same error 
image

yotamleo:
Confirming this fix on the latest **v1.9.2** — the bug still ships there. `hooks/hooks.json` still has the `type: "prompt"` hook under `SessionStart:startup|resume` (and the `PostCompact` prompt hook), so every session start fires:

> Failed to run: prompt-type hooks are not supported for SessionStart events (no conversation context is available). Use a command-type hook instead.

Applied this PR's approach locally against v1.9.2 (preserving the newer `scripts/wiki-lock.sh clear-stale` command hook the SessionStart block has since gained):

```json
"SessionStart": [
  {
    "matcher": "startup|resume",
    "hooks": [
      {
        "type": "command",
        "command": "HOT=\"${CLAUDE_PROJECT_DIR:-$PWD}/wiki/hot.md\"; [ -f \"$HOT\" ] && command -v jq >/dev/null && jq -n --rawfile ctx \"$HOT\" '{hookSpecificOutput:{hookEventName:\"SessionStart\",additionalContext:$ctx}}' || true"
      },
      {
        "type": "command",
        "command": "[ -x scripts/wiki-lock.sh ] && bash scripts/wiki-lock.sh clear-stale --max-age 3600 >/dev/null 2>&1 || true"
      }
    ]
  }
]
```

`PostCompact` removed entirely (no `additionalContext` / injection mechanism there). Banner gone, `wiki/hot.md` injection works. JSON validates with `jq empty`.

Would be great to get this merged — it's the most-reported issue in the tracker (#7, #10, #29, #40, #42, #45, #47, #48, #61 all the same root cause).

AgriciDaniel:

PR #21, opened 2026-04-21
fix(hooks): remove unsupported prompt hook from SessionStart

tphduy:
## Summary

`SessionStart` hooks only accept `type: "command"` (per [Claude Code hook docs](https://code.claude.com/docs/en/hooks)), but `hooks/hooks.json` registers a `type: "prompt"` entry under `SessionStart`. This triggers a `SessionStart:startup hook error` on every new or resumed session.

The preceding `cat wiki/hot.md` command hook already injects the hot cache via stdout, so the prompt entry is both unsupported *and* redundant.

## Fix

Remove the `type: "prompt"` entry from the `SessionStart` block. No functionality is lost.

## Test plan

- [x] Start a new session — no hook error in the banner
- [x] Resume a session — no hook error in the banner
- [x] Confirm `wiki/hot.md` contents are still available in context

## Open question

`PostCompact` in the same file also uses `type: "prompt"`. The docs don't explicitly forbid it there, but worth a maintainer sanity check — happy to follow up in a separate PR if it's also broken.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -7,10 +7,6 @@
           {
             "type": "command",
             "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true"
-          },
-          {
-            "type": "prompt",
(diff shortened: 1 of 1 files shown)


mcpolo99:
#27

PR #26, opened 2026-04-28
fix(scripts): cross-platform file locking, remove flock dependency

mcpolo99:
Replace Unix-only fcntl.flock with a cross-platform Python locking module (scripts/_lock.py) that uses msvcrt.locking on Windows and fcntl.flock on Unix. This makes DragonScale mechanisms 2 and 3 work on Windows/Git Bash without installing flock.

- New: scripts/_lock.py (cross-platform exclusive file lock context manager)
- New: scripts/allocate-address.py (replaces allocate-address.sh)
- New: tests/test_allocate_address.py (Python rewrite of bash tests)
- Removed: scripts/allocate-address.sh (replaced by .py)
- Changed: scripts/tiling-check.py (fcntl.flock replaced with _lock.file_lock)
- Changed: bin/setup-dragonscale.sh (no flock dependency, only Python required)
- Updated: all skill, agent, and doc references from .sh to .py


AgriciDaniel:
Thank you. Since v2.0.0 the mutation path no longer uses `flock(1)` or the per-file `wiki-lock.sh` helper; the canonical `MutationLock` uses Python's `fcntl.flock` on a pinned vault descriptor plus an atomic `mkdir`, macOS runs in CI, and native Windows refuses writes explicitly with `UNSUPPORTED_PLATFORM` (see `docs/windows-wsl.md`). The PostToolUse auto-commit hook is also gone. Closing as fixed in v2.0.0. Address reservation is transaction-owned since v2.0.0, so a second lock authority cannot be merged.

PR #34, opened 2026-05-05
feat(wiki-query): output formats and gap-callout discipline

GoNuck:
## Summary

Codifies the answer shapes that `/wiki-query` was already producing in practice, and tightens gap handling for filed answers.

## What changes

**`skills/wiki-query/SKILL.md`** — three additions, all under existing top-level structure:

1. **New section: Output Formats** — explicit templates for the four answer shapes:
   - **Comparison** (\"A vs B\"): one-liner → trade-off table → judgment frame → common misconceptions
   - **Definition / explanation**: definition → why it matters → related → pitfalls
   - **Inventory** (\"What X do we have?\"): bulleted list grouped by category with wikilinks
   - **Synthesis** (Deep mode): section per dimension, always file back

2. **`> [!gap]` callout convention** — Obsidian-native callout for partial coverage, so gaps become grep-able and visually distinct in the vault rather than buried in prose.

3. **Autoresearch hand-off in Gap Handling** — when a gap is a coherent subtopic worth deep coverage, suggest \`/autoresearch [topic]\` instead of just listing the missing piece. Connects two existing skills that previously didn't reference each other.

## Why

I ran a comparison query (\"RAG と Fine-tuning の使い分け\") manually and produced the trade-off table / judgment frame / misconception structure organically. Comparing what I wrote to the SKILL.md, the answer shape was working but unspecified — so subsequent runs would re-invent it. Codifying it makes results consistent and reduces drift.

The `> [!gap]` callout came from the same exercise: I wanted to mark \"hybrid LoRA+RAG patterns aren't in this wiki\" in the answer, and Obsidian callouts are the obvious idiom but the SKILL.md only described prose-form gap statements.

The autoresearch hand-off is just connecting two existing skills.

## Compatibility

Backwards compatible — purely additive guidance. No existing behavior is changed or removed.

## Version

I left `plugin.json` at 1.6.0; happy for you to bundle this with whatever next release feels right (patch bump seems appropriate).

## Test plan

- [x] Manually verified the new comparison template by running a real query against an existing vault
- [x] Verified the `> [!gap]` callout renders correctly in Obsidian
- [ ] (Maintainer) regression-check against \`/wiki-query\` smoke tests if any

### DIFF
--- a/skills/wiki-query/SKILL.md
+++ b/skills/wiki-query/SKILL.md
@@ -125,6 +125,51 @@ Use sub-indexes when the question is scoped to one domain. Avoid reading the ful
 
 ---
 
+## Output Formats
+
+Pick the shape that fits the question. Default to declarative present tense and wikilink every concept/entity that has a page.
+
+### Comparison ("A vs B", "differences between", "when to use A or B")
+
+Trade-off table is the centerpiece. Then judgment frame, then common misconceptions.
+
+```markdown
+### 一行で
+[One-sentence rule of thumb.]
+
+### 軸ごとの比較 (Source: [[Page A]], [[Page B]])
+
+| 観点 | A | B |
+|---|---|---|
+| 知識更新 | ... | ... |
+| コスト | ... | ... |
+| ... | ... | ... |
+
+### 判断フレーム
+
+1. **「[scenario X]」** → A
+2. **「[scenario Y]」** → B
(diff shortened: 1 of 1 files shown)


AgriciDaniel:
Thanks for the work here. This cannot merge as filed because it conflicts with a product contract: wiki-query is read-only by contract since v2.0.0. Closing; a scoped follow-up is welcome.

PR #46, opened 2026-05-17
docs(wiki): add note on vault root vs content path for Dataview queries

miraclemanza:
"Documents the path-resolution rule that trips up agents creating Dataview queries when content lives in a wiki/ subfolder"

```markdown
## Problem

When the vault is structured with content under a `wiki/` subfolder (per this plugin's standard scaffold), Dataview queries inside notes silently return zero results if they use `FROM "papers"` instead of `FROM "wiki/papers"`. The Dataview engine resolves `FROM` paths relative to the **vault root** (the directory containing `.obsidian/`), not relative to the open note or to the `wiki/` content folder.

This trips up agents (and humans) who assume the `wiki/` content folder IS the vault root. Symptom: every Dataview block in a newly-created synthesis matrix, dashboard, or query note renders `Dataview: No results to show for table/list query` despite the folder containing hundreds of files.

I hit this bug across 4 newly-created vault files in a single session (synthesis-matrix, lit-review-protocol, recipe doc, ingest-candidates doc) before noticing the path-resolution rule. Promoting the rule into `skills/wiki/SKILL.md` makes it active context for any future agent creating Dataview-enabled notes in a vault that follows the standard scaffold this plugin sets up.

## Change

Adds a new section "Vault root vs content path — critical for Dataview" at the end of `skills/wiki/SKILL.md` (after the "When to skip" footer block). The section:

- Explains the path-resolution rule
- Shows the standard vault scaffold tree
- Gives wrong/right examples
- Provides a one-line `ls` verification command
- Names the failure symptom so users can recognize it

No code changes. Pure documentation addition.

## Test plan

- [x] Created a Dataview query in a vault folder structured per the standard scaffold (`<vault-root>/wiki/papers/`) — `FROM "papers"` returns 0 results
- [x] Confirmed `FROM "wiki/papers"` returns expected files
- [x] Verified `.obsidian/` location confirms vault root resolution
```

---

### DIFF
--- a/skills/wiki/SKILL.md
+++ b/skills/wiki/SKILL.md
@@ -232,3 +232,40 @@ Do NOT show the footer after:
 - `obsidian-bases`, `obsidian-markdown` (reference skills, not output)
 - Hot cache updates, index updates, or any background maintenance
 - Error messages or prompts for more information
+ ---
+
+## Vault root vs content path — critical for Dataview
+
+Dataview `FROM "<path>"` is resolved relative to **vault root** (where `.obsidian/` lives), not relative to the open note or to the `wiki/` content folder.
+
+If the vault is structured as:
+
+```
+<vault-root>/         # .obsidian/ lives here
+├── .obsidian/
+├── .raw/
+└── wiki/             # content folder (NOT vault root)
+    ├── papers/
+    ├── Projects/
+    └── meta/
+```
+
+Then Dataview queries inside any note must address subfolders relative to `<vault-root>`:
+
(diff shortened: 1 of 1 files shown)


AgriciDaniel:
Thank you for this. The section this patches was removed in the v2.0.0 skill rewrite, so the diff no longer applies. The rule itself (Dataview queries resolve from the vault root, not the content path) is still correct and undocumented. The natural home now is `skills/wiki/references/plugins.md`, where optional-plugin guidance lives; a short rebased note there, without the emoji markers, would be welcome. Leaving open.

Issue #48, opened 2026-05-18
SessionStart/PostCompact prompt-type hooks rejected by Claude Code v2.1.143


sleggett-ai:
# SessionStart/PostCompact prompt-type hooks rejected by Claude Code v2.1.143

## Summary

`hooks/hooks.json` in v1.6.0 registers `prompt`-type hooks on `SessionStart` and `PostCompact`. Claude Code v2.1.143 rejects the SessionStart one at startup with:

```
Failed to run: prompt-type hooks are not supported for SessionStart events
(no conversation context is available). Use a command-type hook instead.
```

The error surfaces in the TUI banner on every new session. The plugin still functions because the sibling `command`-type hook (`cat wiki/hot.md`) handles the canonical hot-cache load, but the visible error is noisy and suggests the prompt-type entry is dead weight.

## Affected entries

`hooks/hooks.json` lines 11–14 (SessionStart) and 22–26 (PostCompact):

```json
{
  "type": "prompt",
  "prompt": "If a vault is configured ... silently read wiki/hot.md ..."
}
```

PostCompact may or may not be enforced today, but the same rationale applies — Claude Code is tightening prompt-type hook semantics, and a `cat` is functionally equivalent for context injection via STDOUT capture.

## Suggested fix

Replace both prompt-type entries with the same command-type pattern already used on SessionStart:

```json
{ "type": "command", "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true" }
```

This works in non-vault sessions (always exits 0), preserves STDOUT-based context injection, and removes the v2.1.143 error.

## Environment

- Claude Code v2.1.143 (Opus 4.7)
- claude-obsidian v1.6.0 installed via marketplace
- Linux (WSL2 Ubuntu)

## Workaround

Patch the cached `hooks.json` locally — but it's overwritten on plugin update, hence this issue.

AgriciDaniel:
Thank you for the report. This was the prompt-type SessionStart/PostCompact hook shipped through v1.9.2. v2.0.0 (commit 6784d9e) removed both prompt hooks; `hooks/hooks.json` now declares only command hooks (SessionStart with the `startup|resume|clear|compact` matcher, and Stop), and `tests/test_hooks.py` pins that shape. Per the Claude Code hooks documentation, SessionStart plain stdout is added to Claude's context, so no `additionalContext` output is needed.

[...]

PR #49, opened 2026-05-18
fix: replace prompt-type PostCompact hook with command-type

sleggett-ai:
## Summary

- Replaces the `PostCompact` `prompt`-type hook with the same `command`-type pattern already used by `SessionStart` (post-`fc5bb32`).
- Mirrors the rationale of the existing SessionStart fix: Claude Code is tightening prompt-type hook enforcement, and `cat wiki/hot.md` works equivalently via STDOUT capture.

## Context

Filed as #48. Worth noting: `fc5bb32` removed the SessionStart prompt-type hook on `main` but **was never tagged/released** — v1.6.0 in the Claude Code marketplace still ships both prompt-type hooks, so users on the latest release see this error in the TUI banner:

```
```

A v1.6.1 (or similar) release cut after this merge would push both fixes to marketplace users.

## Test plan

- [ ] `python3 -c "import json; json.load(open('hooks/hooks.json'))"` — JSON validates
- [ ] Start a Claude Code session in a vault directory — `wiki/hot.md` still loads via STDOUT
- [ ] Start a Claude Code session in a non-vault directory — no error; hook is a no-op
- [ ] Trigger compaction — `wiki/hot.md` reloads after compact (if vault present)

## Out of scope

- The `Stop` hook is unaffected (was already converted in `e54b419`).
- No changes to `PostToolUse` or other hooks.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -7,10 +7,6 @@
           {
             "type": "command",
             "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true"
-          },
-          {
-            "type": "prompt",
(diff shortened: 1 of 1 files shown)


AgriciDaniel:

PR #60, opened 2026-05-28
fix(detect-transport): use 'obsidian version' subcommand, not '--version' flag

vinsocci:
## Summary

- `scripts/detect-transport.sh` calls `obsidian --version` for the `obsidian` binary path (lines 134, 137). Obsidian's native CLI (1.12+) uses `obsidian version` as a subcommand; `--version` is not a valid flag.
- The unknown flag still returns exit 0 because the CLI prints a helper hint to stdout, so `CLI_PRESENT` is correctly set to `true` — but `version_string` in `transport.json` ends up storing the literal error string: `"Error: Command \"--version\" not found. Did you mean: version?"`.
- Net effect today: `preferred: cli` is still chosen, so this is a cosmetic data-quality fix in `transport.json`, not a transport-selection bug.

## Test plan

- [x] On macOS with Obsidian 1.12.7, confirmed `obsidian version` returns `1.12.7 (installer 1.12.7)` to stdout.
- [x] `bash scripts/detect-transport.sh --peek` before the patch wrote the error string into `version_string`.
- [x] After the patch, `--peek` writes `"version_string": "1.12.7 (installer 1.12.7)"`.
- [x] `preferred: cli` and `fallback_chain: ["cli", "filesystem"]` unchanged across before/after.
- [ ] CI: needs upstream maintainer to run if the project has guarded test workflows.

### DIFF
--- a/scripts/detect-transport.sh
+++ b/scripts/detect-transport.sh
@@ -130,11 +130,11 @@ if command -v obsidian-cli >/dev/null 2>&1; then
   CLI_VERSION="$(printf '%s' "$CLI_VERSION_RAW" | json_escape || echo '"unknown"')"
 elif command -v obsidian >/dev/null 2>&1; then
   # Obsidian 1.12+ ships `obsidian` as the CLI binary on some platforms.
-  # We treat it as cli-capable if it accepts a --cli or --version flag without launching the GUI.
-  if obsidian --version >/dev/null 2>&1; then
+  # We treat it as cli-capable if `obsidian version` returns cleanly without launching the GUI.
+  if obsidian version >/dev/null 2>&1; then
     CLI_PRESENT=true
     CLI_BINARY="obsidian"
-    CLI_VERSION_RAW="$(obsidian --version 2>/dev/null | head -1 || echo unknown)"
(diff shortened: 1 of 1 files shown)


vinsocci:
Real-world confirmation across two machines (macOS, Obsidian 1.12.7, plugin v1.9.2):

**Before (both M4 + M2):**
```
"binary": "obsidian",
"version_string": "Error: Command \"--version\" not found. Did you mean: version?",
"preferred": "cli"
```
Detection still picked `cli` correctly, just stored the helper hint as the version.

**After:**
```
"binary": "obsidian",
"version_string": "1.12.7 (installer 1.12.7)",
"preferred": "cli"
```
`preferred` and `fallback_chain` unchanged.

Setup that exposed it: Obsidian → Settings → General → Advanced → enable Command Line Interface, restart, then `bash scripts/detect-transport.sh --force`. The "did you mean" hint is helpful interactively but trips any caller that pipes the output into JSON.

Thanks for the v1.7→v1.9.2 line — the Compound Vault retrieval story is what motivated the upgrade chain here.

AgriciDaniel:

PR #63, opened 2026-05-29
feat(contextual-prefix): add local ollama prefix tier (tier 2.5)

vinsocci:
## Summary

`contextual-prefix.py` shipped three prefix-generation tiers:

1. **anthropic-api** — `--allow-egress` gated, sends page bodies off-machine (~$12/1k docs)
2. **claude-cli** — `--allow-egress` gated, sends page bodies off-machine (free via CC subscription)
3. **synthetic** — on-machine, template-based, lower quality

For users who already run a local LLM, there was no way to get LLM-quality contextual prefixes **without egress** — a gap between "good but leaves the machine" and "private but weak."

This adds a 4th tier between claude-cli and synthetic:

> **tier 2.5 "ollama"** — POST each chunk to a local ollama `/api/chat` with the page body for context. **Default-on** when ollama is reachable; no flag needed for the common localhost case.

## Egress posture (consistent with the existing model)

- **Local ollama** (`127.0.0.1` / `localhost` / `::1`) needs no flag.
- **Non-localhost** `OLLAMA_URL` requires `--allow-remote-ollama`, mirroring the `scripts/tiling-check.py:351` default-deny precedent.
- `--no-ollama` skips the tier even when reachable.
- **Asymmetric fallback**: on ollama failure the tier drops to **synthetic** and does *not* climb to claude-cli/anthropic-api — egress was never consented to on this path, so silently egressing on local-LLM failure would violate the user's posture.

## API surface

- New flags: `--ollama-model` (default `qwen2.5:7b-instruct`), `--allow-remote-ollama`, `--no-ollama`
- New functions: `ollama_url` / `ollama_is_local` / `ollama_reachable` / `ollama_prefix`
- `pick_prefix_tier()` and `generate_prefix()` extended; the `anthropic-api` and `claude-cli` paths are unchanged.

## Test plan

Verified on macOS (Obsidian 1.12.7, ollama + `qwen2.5:7b-instruct`), all against this repo's own public demo vault:

**Tier picker** (49 demo pages, `--peek`):
| invocation | result |
|---|---|
| default (local ollama up) | all 49 → `tier=ollama` |
| `--no-ollama` | all 49 → `tier=synthetic` |

**Example prefixes** (generated by this branch, `tier=ollama`):
- `wiki/concepts/Search Experience Optimization.md` → *"This chunk outlines the methodology, process, and key innovation of SXO in SEO analysis."*
- `wiki/concepts/Pro Hub Challenge.md` → *"This chunk outlines the Pro Hub Challenge, detailing its structure and rules for both challenges."*
- `wiki/concepts/cherry-picks.md` → *"This chunk outlines Tier 1 features for quick implementation in the Cherry-Picks feature backlog."*

**Production scale**: 895 chunks across 401 pages generated with `tier=ollama`, zero egress, no errors.

## Notes

- Pairs naturally with the existing ollama rerank stage (`rerank.py`), so the full contextual-retrieval pipeline (prefix → BM25 → cosine rerank) can run entirely on-machine.
- Default model `qwen2.5:7b-instruct` is a suggestion, not a requirement — `--ollama-model` accepts any pulled tag (tested also with `qwen2.5-coder:14b`).

### DIFF
--- a/scripts/contextual-prefix.py
+++ b/scripts/contextual-prefix.py
@@ -7,27 +7,33 @@ page, generates a 1-2 sentence prefix situating the chunk in its source. The
 prefixed text is what gets BM25-indexed and embedded, materially improving
 retrieval accuracy (Anthropic measured 35-49% failure reduction).
 
-Three-tier prefix generation (chosen per-run automatically):
-  1. If ANTHROPIC_API_KEY is set      → direct Anthropic API call (Haiku 4.5)
-                                         with prompt caching on the page body
-                                         (only when the body clears the ~16 KB
-                                         Haiku 4.5 cache floor; see
-                                         cache_control_for()).
-                                         ~$12 / 1000 docs per Anthropic figures.
-                                         REQUIRES --allow-egress (sends bodies off-machine).
-  2. Elif `claude` binary on PATH     → `claude -p` subprocess (uses CC subscription;
-                                         no API key needed; slower per call).
-                                         REQUIRES --allow-egress (subprocess egresses).
-  3. Else (default)                   → synthetic prefix from page frontmatter +
-                                         first paragraph (zero-cost floor; loses
-                                         most of the contextual benefit but BM25
-                                         and vector channels still work).
+Four-tier prefix generation (chosen per-run automatically, highest to lowest quality):
+  1. If --allow-egress + ANTHROPIC_API_KEY → direct Anthropic API call (Haiku 4.5)
+                                              with prompt caching on the page body
+                                              (only when the body clears the ~16 KB
+                                              Haiku 4.5 cache floor; see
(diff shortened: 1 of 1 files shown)


AgriciDaniel:
Thank you for this. The egress posture here is right (local-only by default, explicit `--allow-remote-ollama`, fallback to synthetic rather than a consented tier). Two things block review: the diff no longer applies after the contextual-prefix changes on main, so it needs a rebase, and 158 added lines with no hermetic test cannot be merged under CONTRIBUTING.md; an offline test double for the Ollama tier would be enough. Retrieval-quality evidence is gated on the eval harness tracked in #156, so no benchmark is expected from you here. Leaving open.

PR #71, opened 2026-06-01
fix: remove unsupported prompt lifecycle hooks

### LINKED ISSUE #45: SessionStart and PostCompact prompt-type hooks break under Claude Code v2.1.140+
## Summary

`hooks/hooks.json` in v1.6.0 declares `"type": "prompt"` subhooks on both `SessionStart` and `PostCompact` events. Under Claude Code v2.1.140+, the hook contract was tightened — `prompt` is no longer an allowed `type` for `SessionStart` events. The result is a cosmetic but pervasive validation error on every Code session launch inside a configured vault:

```
(or)
additional properties not allowed
```

The plugin continues to function (the surviving `type: command` SessionStart subhook restores vault context fine), but the banner appears on every session start and erodes confidence in the plugin's contract conformance. `PostCompact`'s `type: prompt` block is plausibly subject to the same restriction (would silently error on every `/compact`), though I have not separately verified PostCompact's contract.

## Environment

- Claude Code version: v2.1.140+ (the version that introduced the hook-contract tightening; reproducer verified on a current stable build as of 2026-05-16)
- claude-obsidian version: v1.6.0 (git SHA 75d3b6feb77b96c6bb16599c4550cc9703553d87)
- OS: Windows 11 Home 10.0.26200
- Install method: `/plugin install` via `claude-obsidian-marketplace`
- Install path: `C:\Users\<user>\.claude\plugins\cache\claude-obsidian-marketplace\claude-obsidian\1.6.0\`

## Repro

1. Install `claude-obsidian@1.6.0` via the marketplace.
2. Open a Claude Code session in a directory that contains a configured vault (e.g., a directory with `CLAUDE.md` declaring a `VAULT_PATH` or a `wiki/` subdirectory).
3. Observe the SessionStart banner.

**Expected:** Banner reports clean SessionStart hook execution (the `cat wiki/hot.md` command completes silently or surfaces the hot cache content per the plugin's intent).

**Actual:** Banner reports a hook validation failure — either "prompt-type hooks are not supported for SessionStart events" or "additional properties not allowed" (depending on which surface formats the error). The `cat wiki/hot.md` command subhook still works; only the `type: prompt` subhook is rejected.

## Root cause (proposed)

`hooks/hooks.json` (v1.6.0) contains:

```json
{
  "hooks": {
    "SessionStart": [
      {
        "matcher": "startup|resume",
        "hooks": [
          {
            "type": "command",
            "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true"
          },
          {
            "type": "prompt",
            "prompt": "If a vault is configured for this session (check CLAUDE.md for VAULT_PATH or a wiki/ folder in the current directory), silently read wiki/hot.md to restore recent context. If wiki/hot.md does not exist, do nothing. This is a non-vault session. Do not announce this. Do not report what you read. Just have the context available."
          }
        ]
      }
    ],
    "PostCompact": [
      {
        "matcher": "",
        "hooks": [
          {
            "type": "prompt",
            "prompt": "Hook-injected context does not survive context compaction. If wiki/hot.md exists in the current directory, silently re-read it now to restore the hot cache. Do not announce this."
          }
        ]
      }
    ],
    "PostToolUse": [ ... ],
    "Stop": [ ... ]
  }
}
```

The second SessionStart subhook (`type: prompt`) and the PostCompact subhook (also `type: prompt`) are the offenders. The Code v2.1.140+ contract appears to restrict `SessionStart` (and possibly `PostCompact`) to `type: command` only.

## Proposed fix

Two viable directions — both preserve the intent (vault-context restoration) without invoking a prompt-type hook on a now-prohibited event.

### Option 1 (minimal change): drop the prompt-type subhooks

Remove the `type: prompt` subhook from SessionStart and the entire PostCompact section. The surviving SessionStart `type: command` (`cat wiki/hot.md`) already places the hot cache into the session context — the prompt-type was duplicative, instructing the model to re-read the same file.

This is the patch I applied locally on 2026-05-16 to clear the cosmetic banner. It costs nothing functionally — `cat hot.md` already does the job.

### Option 2 (preserve prompt behavior): migrate to UserPromptSubmit

If you'd like to retain a prompt-based context restoration mechanism (e.g., for cases where `cat`'s injection isn't sufficient or you want the model to actively re-read on every turn), move the prompt-type blocks to `UserPromptSubmit` — that event is still allowed to use `type: prompt` per the same v2.1.140+ contract.

```json
"UserPromptSubmit": [
  {
    "matcher": "",
    "hooks": [
      {
        "type": "prompt",
        "prompt": "If wiki/hot.md exists in the current directory, silently re-read it now to restore the hot cache. Do not announce this."
      }
    ]
  }
]
```

Trade-off: `UserPromptSubmit` fires on every user turn, which is more frequent than SessionStart/PostCompact. If the goal is one-shot context restoration at session/compact boundaries, Option 1 is the better fit.

## Local mitigation (what I'm running)

I'm running a local patch that implements Option 1 (drop both prompt-type blocks). Pre-patch backup retained at `hooks.json.bak-2026-05-16-pre-patch`. Local re-apply runbook lives in my vault — happy to share the exact content if useful.

## Recommendation

Ship Option 1 in the next claude-obsidian release. If you'd like to preserve the prompt-injection pattern, Option 2 to `UserPromptSubmit` works but is a bigger semantic change (per-turn vs per-session). Either way, the v2.1.140+ contract change is upstream-of-you (Anthropic's call) and is going to bite every claude-obsidian user on Code v2.1.140+ — the sooner this is in main, the better.

Happy to send a PR if helpful.

Thanks for the plugin — it's the highest-traffic plugin in my Code stack.

caioribeiroclw-pixel:
## Summary

Removes the prompt-type `SessionStart` and `PostCompact` hooks from `hooks/hooks.json` and updates the hook docs/changelog.

Claude Code v2.1.140+ rejects prompt hooks on these lifecycle events, which creates a validation banner on every configured vault session even though the command-type `SessionStart` hook still restores `wiki/hot.md`.

This follows the Option 1 path proposed in #45: keep the command hot-cache restore path, remove the unsupported prompt blocks, and avoid changing the plugin's per-turn semantics by not moving this to `UserPromptSubmit`.

## Verification

- `python3 -m json.tool hooks/hooks.json >/dev/null`
- `make test`
- `git diff --check`
- Contract smoke: asserted no prompt-type hooks remain under `SessionStart`, and `PostCompact` is no longer present in `hooks/hooks.json`.

Closes #45.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to claude-obsidian. Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Versioning: [SemVer](https://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- Removed unsupported prompt-type `SessionStart` and `PostCompact` plugin hooks from `hooks/hooks.json`. Claude Code v2.1.140+ rejects prompt hooks on these lifecycle events, creating a validation banner on every configured vault session. The command-type `SessionStart` hot-cache hook remains the canonical context restore path.
+
 ## [1.9.2] - 2026-05-27 (prompt-cache hardening + path-handling robustness)
 
 Ports Anthropic prompt-caching best practices into the **one** place the plugin calls the Anthropic API directly: tier-1 contextual-prefix generation in `scripts/contextual-prefix.py`. Verified by full-repo sweep that `cache_control` and the Anthropic API surface exist nowhere else (incl. `claude-canvas/`). No change to retrieval output — API payload shape + observability only.
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -6,8 +6,7 @@ Plugin hooks for the claude-obsidian wiki vault. All hooks are defined in `hooks
 
 | Event | Type | Purpose |
 |---|---|---|
-| `SessionStart` | command + prompt | Loads `wiki/hot.md` into context. Command type runs `[ -f wiki/hot.md ] && cat wiki/hot.md` as the canonical safety check (works for non-vault sessions without erroring). Prompt type complements with semantic context restoration. Matcher: `startup\|resume`. |
-| `PostCompact` | prompt | Re-loads `wiki/hot.md` after context compaction. Hook-injected context does NOT survive compaction (only `CLAUDE.md` does), so this hook restores the hot cache mid-session. |
+| `SessionStart` | command | Loads `wiki/hot.md` into context and clears stale locks. The hot-cache command runs `[ -f wiki/hot.md ] && cat wiki/hot.md` as the canonical safety check (works for non-vault sessions without erroring). Matcher: `startup\|resume`. |
 | `PostToolUse` | command | Auto-commits any wiki/ or .raw/ changes after Write or Edit tool calls. Guarded by `[ -d .git ]` so it never errors in non-git directories, and by `git diff --cached --quiet` so it never creates empty commits. |
 | `Stop` | prompt | Updates `wiki/hot.md` at the end of every Claude response with a brief summary of what changed. |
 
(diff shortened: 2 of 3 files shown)


AgriciDaniel:

PR #74, opened 2026-06-04
fix(detect-transport): Windows process detection via tasklist fallback

EbadSiddiquis:
## Summary

- `pgrep` is unavailable on Windows, so `obsidian_app_running` was always `false` on win32 hosts
- Adds an `elif` branch using `tasklist` (built into every Windows install) as the fallback
- `pgrep` path is unchanged; Linux/macOS behaviour is unaffected

## Test

On Windows with Obsidian running:
```bash
bash scripts/detect-transport.sh --force
# obsidian_app_running: true  ✓
```


AgriciDaniel:
Thanks for the interest in the project. This PR contains vault content, Obsidian plugin binaries, or a personal fork's `main` branch rather than a product change, which the contributing guide excludes from the public tree. Closing without merge; product changes are welcome as focused PRs from a feature branch.

PR #77, opened 2026-06-05
fix(rerank): send nomic search_query/search_document task prefixes

maartengoet:
# fix(rerank): send nomic `search_query:` / `search_document:` task prefixes

## Summary

`nomic-embed-text` is trained with **asymmetric task-instruction prefixes**:
queries must be embedded as `search_query: <text>` and indexed documents as
`search_document: <text>`. The reranker currently embeds both the query and the
chunk text **raw**, with no prefix:

```python
# scripts/rerank.py (before)
q_emb = embed_one(url, DEFAULT_MODEL, query)
...
emb   = embed_one(url, DEFAULT_MODEL, text)
```

Without the prefixes, query and document vectors land in a sub-optimal region of
the embedding space and cosine ranking quality measurably degrades. This is
documented by Nomic as required, not optional:
https://huggingface.co/nomic-ai/nomic-embed-text-v1.5#task-instruction-prefixes

Note: the `prefix` machinery in `contextual-prefix.py` is a *different* concept
(Anthropic contextual-retrieval context sentences) and is unaffected. The BM25
index also intentionally stays prefix-free — prefixes are an embedding-only
concern, so this change is scoped to `rerank.py`.

## What changed

- Added `with_task_prefix(text, role, model)` helper that prepends
  `search_query: ` / `search_document: ` for nomic-family models and is a
  **no-op for non-nomic embedders** (so swapping `DEFAULT_MODEL` to mxbai/bge
  stays correct).
- Applied it at both `embed_one` call sites in `rerank()`.
- Bumped the embed-cache key to include an `EMBED_SCHEME` token
  (`{model}:{EMBED_SCHEME}:{body_hash}`). This **transparently invalidates**
  vectors produced under the old prefix-less convention, so prefixed and
  prefix-less embeddings are never mixed in the same cosine space. No manual
  cache wipe needed — old entries simply miss and get re-embedded once.

## Tests

Added to `tests/test_retrieve.py` (hermetic, no network/ollama):

- `with_task_prefix` applies the correct prefix for query/document
- `with_task_prefix` is a no-op for non-nomic models
- `with_task_prefix` rejects unknown roles
- `rerank()` actually sends `search_query:` / `search_document:` to the embedder
  (captured via a mocked `embed_one`)
- stale prefix-less cache entries are no longer reused after the scheme bump

All existing tests still pass (`make test`).

## Compatibility / risk

- Backwards compatible: behaviour for non-nomic models is unchanged; the only
  one-time cost is a single re-embed of cached chunks on first run after upgrade.
- No API/CLI surface changes.

### DIFF
--- a/scripts/rerank.py
+++ b/scripts/rerank.py
@@ -6,9 +6,12 @@ upstream stage) and reorders them using semantic similarity.
 
 v1.7 strategy (in preference order, automatically chosen at runtime):
   1. If ollama is reachable AND nomic-embed-text is pulled
-       → embed the query, embed each candidate's contextualized_text,
-         rank by cosine. Caches per-chunk embeddings in
-         .vault-meta/embed-cache.json keyed by body_hash.
+       → embed the query (with the "search_query:" task prefix) and each
+         candidate's contextualized_text (with the "search_document:" prefix),
+         rank by cosine. The nomic task prefixes are required for correct
+         asymmetric retrieval; omitting them degrades ranking quality.
+         Caches per-chunk embeddings in .vault-meta/embed-cache.json keyed by
+         body_hash + EMBED_SCHEME (so a convention change invalidates old vectors).
   2. Otherwise
        → no-op rerank: return candidates in input order with a synthesized
          note. Caller (retrieve.py) still gets a useful result; downstream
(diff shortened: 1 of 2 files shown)


AgriciDaniel:
Thank you for this fix. v2.0.0 implemented the same design (the `with_task_prefix` helper, task prefixes, and cache-scheme invalidation), and `ATTRIBUTION.md` now credits it to you as of PR #186. Closing as already applied.

PR #80, opened 2026-06-06
fix(windows): replace fcntl with cross-platform file locking shim

AJIADb9:
Use a small Windows/POSIX lock shim so bm25-index.py, rerank.py, and tiling-check.py run natively on Windows without WSL while preserving existing behavior on Linux and macOS.

# Pull request

## Summary
Three scripts (`bm25-index.py`, `rerank.py`, `tiling-check.py`) imported `fcntl` unconditionally, which crashes on Windows with `ModuleNotFoundError`. This PR switches them to a small platform shim: `msvcrt.locking()` on Windows and `fcntl.flock()` elsewhere, with lock mode chosen per script to match how each helper is used.

## Type
- [x] Bug fix (fix:)

## Related issue
N/A

## Changes
- `scripts/bm25-index.py` -- import `os` before the Windows shim and use `_flock_ex` / `_flock_un`
- `scripts/rerank.py` -- same, plus translate Windows sharing/lock violations to `BlockingIOError` so the existing retry loop still works without masking unrelated failures
- `scripts/tiling-check.py` -- same, using blocking `msvcrt.LK_LOCK` for cache locking on Windows

## Six-cut self-review
- [x] New identifiers named for the next reader (_flock_ex, _flock_un)
- [x] Smallest unit that works (no speculative abstraction)
- [x] Deletions kept up with additions (Unix-only imports removed where needed; shim added in place)
- [ ] New behavior has hermetic test coverage -- no new tests added
- [x] New failure modes have explicit handling -- `rerank.py` translates only Windows sharing/lock violations into `BlockingIOError` for the retry loop, and `tiling-check.py` uses blocking Windows locking for cache access

## Testing
Targeted checks run:

    > uv run python tests/test_bm25_index.py
    > uv run python tests/test_retrieve.py
    > uv run python tests/test_tiling_check.py
    > ruff check scripts/bm25-index.py scripts/rerank.py scripts/tiling-check.py

## Verifier
Not dispatched -- change is a 3-file mechanical substitution with no logic change.

## CHANGELOG
- [ ] Added an entry under ## [Unreleased] in CHANGELOG.md -- not added, leave to maintainer

## Screenshots / output
Before (Windows): `ModuleNotFoundError: No module named 'fcntl'`
After (Windows): all three scripts import and run with Windows-compatible locking

## Notes for reviewer
- `bm25-index.py` uses a non-blocking exclusive lock.
- `rerank.py` retries on lock contention by treating only Windows sharing/lock violations as `BlockingIOError`.
- `tiling-check.py` uses blocking `msvcrt.LK_LOCK` for cache locking on Windows.


AgriciDaniel:
Thank you. Since v2.0.0 the mutation path no longer uses `flock(1)` or the per-file `wiki-lock.sh` helper; the canonical `MutationLock` uses Python's `fcntl.flock` on a pinned vault descriptor plus an atomic `mkdir`, macOS runs in CI, and native Windows refuses writes explicitly with `UNSUPPORTED_PLATFORM` (see `docs/windows-wsl.md`). The PostToolUse auto-commit hook is also gone. Closing as fixed in v2.0.0. The advertised `msvcrt` shim is not present in the diff; if native Windows writes land via #151, a fresh single-purpose PR would be welcome.

PR #82, opened 2026-06-09
feat: Mode B code-ingestion + /wiki-code-lint code-fidelity audit

RasenGUY:
## What this adds

This PR brings the **Mode B code-ingestion family** to upstream and adds a new
**`/wiki-code-lint`** skill that keeps a code-architecture wiki honest as the code moves.

Upstream currently lacks `wiki-code-ingest` / `code-anchor-check.py`, so the three commits
are self-contained (they add the machinery *and* the new lint that builds on it).

### Commits
1. **`feat: add Mode B codebase ingestion, drift lint, and commit sync`** — the `wiki-code-ingest`
   skill (repo → modules/flows/dependencies/decisions pages with git drift anchors),
   `wiki-code-watch` (git-hook sync), and the deterministic signal scripts.
2. **`fix(code-anchor-check): match abbreviated and inline code_anchors`** — drift comparison now
   prefix-matches abbreviated SHAs (ingest writes 12-char anchors; `git rev-parse` returns 40), and
   parses inline `code_anchors: [...]` as a list instead of iterating it character-by-character.
   Without this, the drift check reported every page on every real vault as drifted/malformed.
3. **`feat(wiki-code-lint): add Mode B code-fidelity lint and emit aliases`** — the new skill.

### `/wiki-code-lint`
Read-only, the `wiki-code-*` counterpart to `/wiki-lint`. Four checks against the repo a Mode B
wiki was ingested from, written to a dated `code-lint-report-YYYY-MM-DD.md`:

- **Code drift** — anchored source SHAs vs HEAD (reuses `code-anchor-check.py`).
- **Ingest staleness** — commits/days the wiki is behind HEAD; orphaned/missing `ingest_commit`.
- **Coverage gaps** — repo packages with no wiki page, at the levels the wiki already documents.
- **Link resolution** — models Obsidian's real resolver (filename ∪ alias, **not** `title:`), catching
  the trap where slug filenames without `aliases:` silently break every `[[Title]]` link.

To stop that trap recurring, `wiki-code-ingest` (and its sub-agent + the frontmatter reference) now
**emit `aliases: ["<title>"]`** on every code page.

### Tests
Two new hermetic suites (`test_code_coverage_check.py`, `test_wiki_link_resolve_check.py`) plus two
regression tests in `test_code_anchor_check.py` for the abbreviated/inline fixes. All green, no network/LLM.

Validated end-to-end against a real 54-page code wiki: 1 true-positive drift, 12 undocumented
packages surfaced, +1-commit staleness, and one genuinely unresolvable `/`-in-title link.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,3 +132,14 @@ _attachments/
 # v1.8 methodology mode (host-specific by default; `git add -f` to commit if desired)
 .vault-meta/mode.json
 .vault-meta/mode.*.tmp
+
+# Mode B code-watch runtime artifacts (regenerable / host-specific: paths + SHAs
+# are local to this machine). Must be ignored — the PostToolUse hook git-adds
+# .vault-meta/ and .raw/ after every tool use, so without these rules an
+# auto-commit loop would commit the queue/state/locks on every interaction.
+.vault-meta/code-sync-queue.jsonl
+.vault-meta/code-sync-state.json
+.vault-meta/.code-sync*.lock
+.vault-meta/.code-sync-launch.lock.d/
+# Deterministic code-signal dumps, regenerated on every ingest (cache, not source)
+.raw/code/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 All notable changes to claude-obsidian. Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Versioning: [SemVer](https://semver.org/).
 
+## [Unreleased]
+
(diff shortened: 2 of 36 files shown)


AgriciDaniel:
Thank you for a substantial contribution. This branch targets the shell-script era that v2.0.0 replaced (command mirrors, `allowed-tools` frontmatter, per-file locks, direct worker writes), so it cannot be rebased. Closing the PR; the series is evaluated at #85; the drift-anchor idea is tracked as a roadmap item for the lint engine.

Issue #88, opened 2026-06-11
SessionStart hook: prompt-type entry now raises "use a command-type hook instead" error in recent Claude Code


alienpenguin10:
## Summary

`hooks/hooks.json` declares a `type: "prompt"` hook under `SessionStart` (matcher `startup|resume`). Recent Claude Code releases made this an error — every session start now prints:

```
SessionStart:startup hook error
Failed to run: prompt-type hooks are not supported for SessionStart events
(no conversation context is available). Use a command-type hook instead.
```

Affects v1.9.2 (and presumably earlier versions that ship the same `hooks.json`).

## Reproduction

1. Install/enable the plugin in `~/.claude/settings.json`.
2. Start any Claude Code session.
3. Observe the error banner at startup. Non-fatal, but noisy.

## Root cause

`hooks/hooks.json` lines 15–18:

\`\`\`json
{
  "type": "prompt",
  "prompt": "If a vault is configured for this session ..."
}
\`\`\`

Per the Claude Code changelog:

> Improved hook configuration error: configuring a prompt- or agent-type hook for \`SessionStart\` / \`Setup\` / \`SubagentStart\` now shows a clear \"use a command-type hook instead\" error.

The rationale: at `SessionStart` there's no conversation yet, so a prompt has nowhere to land. The plugin's own `hooks/README.md` already notes that the command-type `cat wiki/hot.md` hook is the canonical safety check, so the prompt-type entry was a defensive duplicate that no longer functions.

## Suggested fix

Remove the prompt-type entry from the `SessionStart` block. Keep the two command-type hooks — they continue to work and cover the canonical hot-cache restore + stale-lock reaper paths. If hot-cache restore mid-session is still wanted via prompt injection, `PostCompact` and `UserPromptSubmit` remain valid carriers; `SessionStart` is not.

I'll happily PR this if useful.

AgriciDaniel:
Thank you for the report. This was the prompt-type SessionStart/PostCompact hook shipped through v1.9.2. v2.0.0 (commit 6784d9e) removed both prompt hooks; `hooks/hooks.json` now declares only command hooks (SessionStart with the `startup|resume|clear|compact` matcher, and Stop), and `tests/test_hooks.py` pins that shape. Per the Claude Code hooks documentation, SessionStart plain stdout is added to Claude's context, so no `additionalContext` output is needed.

[...]

Issue #87, opened 2026-06-11
SessionStart prompt-type hook fails: "prompt-type hooks are not supported for SessionStart events"


Luisalfredotech:
**Description**

In `hooks/hooks.json` (v1.9.2), the `SessionStart` hook group contains a `"type": "prompt"` hook. Claude Code rejects prompt-type hooks for SessionStart, so every session starts with this error:

```
Failed to run: prompt-type hooks are not supported for SessionStart events (no conversation context is available). Use a command-type hook instead.
```

**Offending block**

```json
{
  "type": "prompt",
  "prompt": "If a vault is configured for this session (check CLAUDE.md for VAULT_PATH or a wiki/ folder in the current directory), silently read wiki/hot.md to restore recent context. If wiki/hot.md does not exist, do nothing. This is a non-vault session. Do not announce this. Do not report what you read. Just have the context available."
}
```

**Why it's redundant**

The same SessionStart group already has a command-type hook that injects the hot cache directly into context:

```json
{
  "type": "command",
  "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true"
}
```

The `cat` output is delivered as SessionStart context, so the model already has `wiki/hot.md` available without being asked to read it. The prompt hook adds nothing — it only produces the startup error.

**Proposed fix**

Remove the prompt-type hook from the `SessionStart` group. Optionally harden the cat hook so it works regardless of cwd and signals the missing-file case:

```json
{
  "type": "command",
  "command": "cat \"$CLAUDE_PROJECT_DIR/wiki/hot.md\" 2>/dev/null || echo \"No hot memory file\""
}
```

Note: the `PostCompact` prompt-type hook is fine — that event supports prompt hooks. Only SessionStart is affected.

**Environment:** claude-obsidian 1.9.2, Claude Code CLI, macOS.

rjlasko:
Same issue here. v2.1.183 of `claude`

AgriciDaniel:
Thank you for the report. This was the prompt-type SessionStart/PostCompact hook shipped through v1.9.2. v2.0.0 (commit 6784d9e) removed both prompt hooks; `hooks/hooks.json` now declares only command hooks (SessionStart with the `startup|resume|clear|compact` matcher, and Stop), and `tests/test_hooks.py` pins that shape. Per the Claude Code hooks documentation, SessionStart plain stdout is added to Claude's context, so no `additionalContext` output is needed.

[...]

Issue #61, opened 2026-05-28
SessionStart hook uses unsupported type: "prompt" — fails on resume


ecik666:
## Summary

The `SessionStart` hook in `hooks/hooks.json` includes a `"type": "prompt"` entry. Claude Code does not support prompt-type hooks for `SessionStart` events (there is no conversation context available at session start), so it raises an error on every startup/resume:

> SessionStart:resume hook error
> Failed to run: prompt-type hooks are not supported for SessionStart events (no conversation context is available). Use a command-type hook instead.

## Location

`hooks/hooks.json`, inside the `SessionStart` block (matcher `startup|resume`):

```json
{
  "type": "prompt",
  "prompt": "If a vault is configured for this session (check CLAUDE.md for VAULT_PATH or a wiki/ folder in the current directory), silently read wiki/hot.md to restore recent context. If wiki/hot.md does not exist, do nothing. This is a non-vault session. Do not announce this. Do not report what you read. Just have the context available."
}
```

## Why it's safe to remove

This prompt hook is redundant. The first command hook in the same block already injects the hot cache directly:

```json
{ "type": "command", "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true" }
```

So the hot-cache content is already in context at session start without the prompt hook.

## Suggested fix

Either:

1. **Remove** the `type: "prompt"` entry from `SessionStart` (simplest — the command hook already covers it), or
2. **Move** that logic to the `PostCompact` block, which *does* support prompt-type hooks (the repo already has a similar prompt hook there).

## Environment

- Plugin version: 1.9.2
- Claude Code (recent build; SessionStart prompt-type hooks rejected)
- Platform: macOS

## Reproduction

1. Install/enable the plugin.
2. Start or resume a Claude Code session.
3. Error appears in the SessionStart hook output.

chenthedigger:
Still reproduces on the marketplace release **v1.6.0**, and the `type: "prompt"` entry is also still present on **`main`** (current `hooks/hooks.json` → `SessionStart`, plus the same pattern under `PostCompact`) — so it's not yet resolved in the shipped or HEAD version.

> Failed to run: prompt-type hooks are not supported for SessionStart events (no conversation context is available). Use a command-type hook instead.

[...]

Issue #47, opened 2026-05-17
SessionStart prompt-type hook fails: "no conversation context is available"


fracampit:
## Bug

On Claude Code session start in a vault directory, the plugin's `SessionStart` hook emits this error:

```
Failed to run: prompt-type hooks are not supported for SessionStart events
(no conversation context is available). Use a command-type hook instead.
```

The error is non-fatal — the session continues — but it surfaces at every startup and the prompt hook is skipped.

**Plugin version:** 1.6.0
**OS:** Windows 11

## Root cause

`hooks/hooks.json` registers a `prompt`-type hook under `SessionStart`. Claude Code's `SessionStart` event fires before any conversation context exists, so prompt-type hooks can't be processed there. The `command`-type hook in the same entry works fine and emits `wiki/hot.md` content.

## Suggested fix

Drop the redundant prompt hook from `SessionStart` — the command hook already covers the functional case (emitting hot.md content at session start):

```diff
 "SessionStart": [
   {
     "matcher": "startup|resume",
     "hooks": [
       {
         "type": "command",
         "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true"
-      },
-      {
-        "type": "prompt",
-        "prompt": "If a vault is configured for this session..."
       }
     ]
   }
 ]
```

## Related — `PostCompact`

`PostCompact` has the same prompt-only structure and likely exhibits the same error on context compaction (not yet confirmed in practice). Suggest converting it to a command-type that outputs `wiki/hot.md` content like `SessionStart` does:

```json
"PostCompact": [
  {
    "matcher": "",
    "hooks": [
      {
        "type": "command",
        "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true"
      }
    ]
  }
]
```

## Workaround

Local patch to `~/.claude/plugins/cache/claude-obsidian-marketplace/claude-obsidian/<version>/hooks/hooks.json`. Wiped on plugin updates.

AgriciDaniel:
Thank you for the report. This was the prompt-type SessionStart/PostCompact hook shipped through v1.9.2. v2.0.0 (commit 6784d9e) removed both prompt hooks; `hooks/hooks.json` now declares only command hooks (SessionStart with the `startup|resume|clear|compact` matcher, and Stop), and `tests/test_hooks.py` pins that shape. Per the Claude Code hooks documentation, SessionStart plain stdout is added to Claude's context, so no `additionalContext` output is needed.

[...]

PR #90, opened 2026-06-11
fix(hooks): stop SessionStart/PostCompact prompt hooks from erroring every session

nathanm4:
## Summary

`hooks/hooks.json` registers `prompt`-type hooks on two lifecycle events — `SessionStart` (matcher `startup|resume`) and `PostCompact`. Recent Claude Code releases reject `prompt`-type hooks on lifecycle events that have no conversation context to inject into, so **every** startup/resume now prints:

```
```

Reproduced on a clean install of **v1.9.2** with **Claude Code v2.1.173**. Non-fatal, but it fires on every session and is alarming to users (see the cluster of reports: #88, #87, #61, #48, #47, #45, #40, #29, #7).

## Fix

Both lifecycle hooks become `command`-type, so no `prompt` hooks remain on any lifecycle event:

| Event | Before | After | Rationale |
|---|---|---|---|
| `SessionStart` | command + command + **prompt** | command + command | The `prompt` step was a redundant duplicate — the `[ -f wiki/hot.md ] && cat wiki/hot.md` command hook already restores the hot cache, and `hooks/README.md` itself calls that the "canonical safety check." Removing it loses nothing. |
| `PostCompact` | **prompt** | command | **Converted, not deleted.** The intent (re-read `wiki/hot.md` after compaction) is real and has no existing command-hook equivalent, so I replaced it with `[ -f wiki/hot.md ] && cat wiki/hot.md` — the same STDOUT-capture restore path `SessionStart` already uses — rather than dropping the behavior. |

`hooks/README.md` (events table + STDOUT-bug note) and `CHANGELOG.md` are updated to match. The `Stop` hook is untouched (it's already command-type and fires with conversation context).

## Why convert `PostCompact` instead of removing it

Hook-injected context does not survive compaction, so `PostCompact` is the one place the plugin re-hydrates the hot cache mid-session. Simply deleting it silently removes that behavior. Converting it to the command form preserves it using a supported hook type, consistent with the existing `SessionStart` command hook. It remains subject to the pre-existing plugin-STDOUT caveat already documented in `hooks/README.md` (`anthropics/claude-code#10875`) — this PR doesn't change that, and notes it honestly in the changelog.

## Verification

```
$ jq empty hooks/hooks.json && echo "valid JSON"
valid JSON

# No prompt-type hooks remain on any lifecycle event:
$ jq -r '.hooks | to_entries[] | "\(.key): \([.value[].hooks[].type] | unique | join(", "))"' hooks/hooks.json
SessionStart: command
PostCompact: command
PostToolUse: command
Stop: command
```

After applying locally, the `SessionStart:startup hook error` banner no longer appears on startup or resume.

## Notes

This overlaps with #68 (SessionStart only) and #71 (removes both, but deletes the `PostCompact` behavior). This PR's distinction is preserving the `PostCompact` restore via conversion rather than deletion, plus keeping the docs/changelog in sync. Happy to rebase or trim scope to whatever you'd prefer to merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 All notable changes to claude-obsidian. Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Versioning: [SemVer](https://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- **`SessionStart` / `PostCompact` hooks no longer error on every session** (`hooks/hooks.json`). Recent Claude Code releases reject `prompt`-type hooks on lifecycle events that have no conversation context to inject into, printing `SessionStart:startup hook error — prompt-type hooks are not supported for SessionStart events (no conversation context is available). Use a command-type hook instead.` on every startup/resume.
+  - `SessionStart`: removed the redundant `prompt` step. The command hook `[ -f wiki/hot.md ] && cat wiki/hot.md` already restores the hot cache (the README's documented "canonical safety check"), so no behavior is lost.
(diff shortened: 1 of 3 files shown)


AgriciDaniel:

PR #99, opened 2026-06-20
Fix SessionStart hook: remove unsupported prompt-type hook

k-2-corp:
## Problem

`hooks/hooks.json` registers a `prompt`-type hook under `SessionStart`. Claude Code rejects it on every new session / resume:

> Failed to run: prompt-type hooks are not supported for SessionStart events (no conversation context is available). Use a command-type hook instead.

`SessionStart` fires before any conversation context exists, so only `command`-type hooks are valid there.

## Why it is safe

The prompt-hook is redundant. The `command`-hook directly above already streams the hot cache into context:

```json
```

So `cat wiki/hot.md` already restores context; the prompt-hook just asked the model to read the same file again.

## Change

Remove the `prompt` block from the `SessionStart` array only. `PostCompact` keeps its prompt-hook (valid there). JSON validated.

Fixes the per-session warning seen on v1.9.2.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -11,10 +11,6 @@
           {
             "type": "command",
             "command": "[ -x scripts/wiki-lock.sh ] && bash scripts/wiki-lock.sh clear-stale --max-age 3600 >/dev/null 2>&1 || true"
-          },
-          {
-            "type": "prompt",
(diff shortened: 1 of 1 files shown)


AgriciDaniel:

PR #109, opened 2026-07-02
feat(perimeter): opt-in privacy perimeter + graph lint + consolidation guide

HeroiDeComic:
Implements https://github.com/AgriciDaniel/claude-obsidian/issues/108 — privacy perimeter + graph lint + consolidation guide.

## What

Five reusable patterns from consolidating four live vaults (~10k files) into one air-gapped vault. Two land as opt-in tools, three as a methodology guide:

- **`bin/setup-perimeter.sh`** — opt-in privacy perimeter. Installs a git `pre-commit` hook (blocks staged paths matching sensitive-band globs; scans ADDED diff lines against an identifier/secret regex list, minus a false-positive whitelist) and a `pre-push` hook (optional `--air-gap` blocks every push). Config is three plain-text files under `.vault-meta/` seeded with commented examples. Consent prompt (or `--yes`), `--check` diagnostics, `--uninstall` restores any pre-existing hooks from backup. Everything runs locally; nothing leaves the machine.
- **`scripts/graph-lint.py`** — wikilink graph health check. Resolves links the way Obsidian does (basename + frontmatter aliases, inline `[a, b]` and block `- a` styles, path-style links by last segment). Reports broken links, broken `up:` hierarchy fields, and orphans (content vs `_system`). Legitimate dangling links (calendar dates, planned pages) are accepted via a whitelist regex file and counted separately. `--json` report, `--strict` exit code for CI/gates. Stdlib only.
- **`docs/consolidation-guide.md`** — the methodology the tools serve: the per-source consolidation loop, hardened migration for sensitive sources (dry-run default, manifest proof, scan-before-add, index-only tier, anti-operator brakes), and merge governance (dedup ledger + pending-decisions quarantine).
- **`tests/test_graph_lint.py`** (12 assertions) + **`tests/test_setup_perimeter.sh`** (17 assertions) — hermetic, no network, wired into `make test` via two new Makefile targets.

## Why

The README's promise is "keeps your data yours". For vaults holding data the owner would never publish, that needs a deny-by-default boundary at the git layer — regex scanning alone provably misses real data disguised as "examples". And once a vault grows past a few hundred pages (or merges several sources), link-graph rot is invisible without a whole-vault pass that can tell *broken* apart from *legitimately dangling*.

## Testing

- `python3 tests/test_graph_lint.py` → 12 passed, 0 failed
- `bash tests/test_setup_perimeter.sh` → 17 passed, 0 failed
- Both suites are hermetic (mktemp sandboxes, stdlib/POSIX utilities only, no network).
- Developed and verified on Windows git-bash. POSIX portability explicitly reviewed (no bashisms beyond `#!/usr/bin/env bash`, `[[:space:]]` instead of `\s`, standard `grep -E`/`sed`/`paste` usage); a Linux/macOS `make test` run before merge is welcome.
- Pre-commit verifier agent (`agents/verifier.md`) run on the staged diff before commit; findings addressed.

## Blast radius / undo

- Both tools are new and opt-in; no existing file's behavior changes (Makefile only appends targets).
- The perimeter writes to `.git/hooks/` (untracked, per-clone): managed hooks carry a marker line; pre-existing unmanaged hooks are backed up to `<hook>.pre-perimeter.bak` and restored on `--uninstall`. Undo = `bash bin/setup-perimeter.sh --uninstall`.
- Failure modes handled: not-a-git-repo (exit 3), empty/missing config files (hook degrades to secrets baseline), bad whitelist regex (skipped with WARN), missing scan root (exit 2).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,3 +132,9 @@ _attachments/
 # v1.8 methodology mode (host-specific by default; `git add -f` to commit if desired)
 .vault-meta/mode.json
 .vault-meta/mode.*.tmp
+# Privacy perimeter config (user's sensitive-band globs and PII regex
+# vocabulary are themselves sensitive; `git add -f` to share with a team)
+.vault-meta/perimeter-paths.txt
+.vault-meta/perimeter-patterns.txt
+.vault-meta/perimeter-whitelist.txt
+.vault-meta/perimeter-airgap
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,9 @@
 # Test runner entry points for DragonScale and vault tooling.
 
 .PHONY: test test-address test-tiling test-boundary test-bm25 test-retrieve \
-        test-lock test-concurrent test-mode test-contextual setup-dragonscale \
-        setup-retrieve setup-mode clean-test-state help
+        test-lock test-concurrent test-mode test-contextual test-graph-lint \
+        test-perimeter setup-dragonscale \
+        setup-retrieve setup-mode setup-perimeter clean-test-state help
 
 help:
 	@echo "claude-obsidian developer targets:"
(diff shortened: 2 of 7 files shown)


AgriciDaniel:
Thank you for this. The graph-lint half is already on main since v2.0.0 (Obsidian-accurate alias resolution, dead-link and orphan reporting, and an accepted-dangling allowlist live in `claude_obsidian/lint_engine.py`), so `scripts/graph-lint.py` and its test would duplicate it. The commit-time privacy perimeter and the consolidation guide have no counterpart on main and are welcome. Could you resubmit with only `setup-perimeter.sh`, its test, and the guide, rebased on the Python core (note the setup scripts moved to `scripts/` in PR #186)? Leaving open.

PR #114, opened 2026-07-07
Windows/Git Bash portability: flock + fcntl fallbacks

hvacsidekick:
## Problem

The lock primitives assume POSIX `flock(1)` / the Python `fcntl` module, neither of which exists on **Windows Git Bash** (msys2 ships no util-linux `flock`; CPython on Windows has no `fcntl` module). On Windows this makes three core scripts crash:

- `scripts/wiki-lock.sh` → `flock: command not found` in `with_meta_lock()` — **all** lock acquire/list/release fail, so multi-writer safety is unavailable.
- `scripts/bm25-index.py` → `ModuleNotFoundError: No module named 'fcntl'` at import — index build dies.
- `scripts/rerank.py` → same `fcntl` import error — `retrieve.py` fails to import the sibling and exits 10.

Net effect: multi-writer locking and the **entire wiki-retrieve pipeline** are non-functional on Windows.

## Fix

Portable fallbacks that keep the POSIX path byte-for-byte unchanged:

| File | Change |
|------|--------|
| `wiki-lock.sh` | If `flock` is absent, `with_meta_lock()` falls back to an **atomic `mkdir` spinlock** (mkdir is atomic on POSIX/NTFS), preserving the 5s timeout and stale-reap semantics. `flock` path unchanged when present. |
| `bm25-index.py` | Guard the `fcntl` import; on Windows `acquire_lock()` uses an atomic **`O_EXCL` lockfile** (non-blocking exclusive) with a 60s stale reap. |
| `rerank.py` | Guard the `fcntl` import; on Windows fall through to the **existing atomic temp+rename** write (the WARN is suppressed since running unlocked is the expected path there). |

The page-level lock in `wiki-lock.sh` was already flock-free (atomic `noclobber` create); only the internal meta-lock needed the fallback.

## Validation

On **Windows 11 / Git Bash 5.2 / Python 3.12**:

- `wiki-lock.sh`: a 5-way concurrent `acquire` race on one page yields **exactly one winner**; acquire/list/peek/release/clear-stale all behave.
- `bm25-index.py build` + `retrieve.py "<query>"` return correct BM25 results (rerank cleanly no-ops without ollama).

No behavior change on Linux/macOS — all three fall back only when the POSIX primitive is missing.

### DIFF
--- a/scripts/bm25-index.py
+++ b/scripts/bm25-index.py
@@ -46,7 +46,6 @@ Exit codes:
 """
 
 import argparse
-import fcntl
 import json
 import math
 import os
@@ -55,6 +54,13 @@ import sys
 from collections import Counter, defaultdict
 from datetime import datetime, timezone
 from pathlib import Path
+
+try:
+    import fcntl  # POSIX only
+    _HAVE_FCNTL = True
+except ImportError:  # Windows (Git Bash Python) ships no fcntl module
+    fcntl = None
+    _HAVE_FCNTL = False
 
 VAULT_ROOT = Path(__file__).resolve().parent.parent
 META_DIR = VAULT_ROOT / ".vault-meta"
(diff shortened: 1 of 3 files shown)


AgriciDaniel:

PR #118, opened 2026-07-12
Portable locks: make DragonScale tooling work on Windows (Git Bash)

### LINKED ISSUE #117: DragonScale tooling broken on Windows (Git Bash): flock and fcntl don't exist
On Windows, Git Bash ships without `flock(1)` and Python has no `fcntl` module, which breaks all three DragonScale locking scripts:

- `scripts/allocate-address.sh` exits 1 on every run (`flock: command not found`) → **no ingest can ever assign addresses on Windows**; every post-rollout page ends up without an `address:` field.
- `scripts/wiki-lock.sh` fails inside `with_meta_lock` for the same reason, so acquire/release/list/clear-stale are all unusable.
- `scripts/tiling-check.py` crashes at `import fcntl` instead of exiting with its documented feature-gate codes (10/11), so the calling skill cannot no-op gracefully.

**Repro:** stock Windows 11 + Git for Windows (no WSL), run `bash scripts/allocate-address.sh` → exit 1.

**Proposed fix** (PR incoming from my fork): keep the existing flock/fcntl paths on POSIX and add portable fallbacks on Windows — an atomic `mkdir` lock in the two shell scripts, and `msvcrt.locking` in the Python script.

One test-harness fix included: in MSYS bash, a `$'\r'` literal passed directly as an argument to a child process **inside command substitution** silently loses the CR, so the carriage-return-rejection test in `tests/test_wiki_lock.sh` could never actually exercise the rejection on Windows. Building the path in a variable first preserves the CR on both platforms.

With the patch, the full test suite is green on Windows 11 (Git Bash): test_wiki_lock 16/16, test_allocate_address 12/12, test_concurrent_write all pass, test_tiling_check all pass. POSIX behavior is unchanged (the flock/fcntl branches are untouched).

Thebigboss537:
Fixes #117.

On Windows, Git Bash has no `flock(1)` and Python has no `fcntl`, so address allocation, wiki locking and tiling checks are all broken (details in the issue).

## Changes

- **`scripts/allocate-address.sh`** — if `flock` is unavailable, fall back to an atomic `mkdir` lock (`$LOCK_FILE.d`) with the same 5s timeout semantics; cleanup via `trap EXIT`. The flock path is untouched.
- **`scripts/wiki-lock.sh`** — same mkdir fallback inside `with_meta_lock`. The inner command runs in a subshell so an inner `die()`/`exit` can't skip the lock cleanup (mirrors how the flock branch releases on subshell exit).
- **`scripts/tiling-check.py`** — `import fcntl` wrapped in try/except; on Windows `_lock_cache`/`_unlock_cache` use `msvcrt.locking` on byte 0 of the lock file, giving the same mutual exclusion as `flock LOCK_EX`. Documented exit codes (10/11) now work on Windows.
- **`tests/test_wiki_lock.sh`** — MSYS quirk: a `$'\r'` literal passed directly as an argument to a child process inside command substitution loses the CR before it reaches the child, so the CR-rejection test never actually tested anything on Windows (and left a stray lock that broke the stress-count assertion). Building the path in a variable first preserves the CR on both platforms.

## Testing

- Windows 11 + Git for Windows (no flock): `test_wiki_lock.sh` 16/16, `test_allocate_address.sh` 12/12, `test_concurrent_write.sh` all pass, `test_tiling_check.py` all pass. Manually verified `allocate-address.sh --peek/--rebuild/allocate`, `wiki-lock.sh acquire/release/list/clear-stale/peek` and `tiling-check.py --peek` (exit 10 without ollama, as documented).
- POSIX: flock/fcntl branches unchanged; suite unaffected.

### DIFF
--- a/scripts/allocate-address.sh
+++ b/scripts/allocate-address.sh
@@ -32,10 +32,26 @@ mkdir -p "$(dirname "$COUNTER_FILE")" || {
 }
 
 # Acquire exclusive lock with 5-second timeout. Release automatically on scope exit.
-exec 9>"$LOCK_FILE"
-if ! flock -x -w 5 9; then
-  echo "ERR: could not acquire address allocator lock within 5s" >&2
-  exit 1
+# flock(1) is unavailable on Windows Git Bash / native msys; fall back to an
+# atomic mkdir lock (mkdir is atomic on every platform bash runs on).
+if command -v flock >/dev/null 2>&1; then
+  exec 9>"$LOCK_FILE"
+  if ! flock -x -w 5 9; then
+    echo "ERR: could not acquire address allocator lock within 5s" >&2
+    exit 1
+  fi
+else
+  LOCK_DIR_FALLBACK="${LOCK_FILE}.d"
+  acquired=0
+  for _ in $(seq 1 50); do
+    if mkdir "$LOCK_DIR_FALLBACK" 2>/dev/null; then acquired=1; break; fi
+    sleep 0.1
+  done
+  if [ "$acquired" -ne 1 ]; then
+    echo "ERR: could not acquire address allocator lock within 5s" >&2
+    exit 1
+  fi
+  trap 'rmdir "$LOCK_DIR_FALLBACK" 2>/dev/null || true' EXIT
 fi
 
 scan_max_c_address() {
--- a/scripts/tiling-check.py
+++ b/scripts/tiling-check.py
@@ -27,7 +27,12 @@ Usage:
 """
 
 import argparse
-import fcntl
+
+try:
+    import fcntl  # POSIX advisory locks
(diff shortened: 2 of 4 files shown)


AgriciDaniel:
Thank you. Since v2.0.0 the mutation path no longer uses `flock(1)` or the per-file `wiki-lock.sh` helper; the canonical `MutationLock` uses Python's `fcntl.flock` on a pinned vault descriptor plus an atomic `mkdir`, macOS runs in CI, and native Windows refuses writes explicitly with `UNSUPPORTED_PLATFORM` (see `docs/windows-wsl.md`). The PostToolUse auto-commit hook is also gone. Closing as fixed in v2.0.0. This was the best of the Windows group; the `mkdir` mutex design is what main converged on.

PR #121, opened 2026-07-12
fix(macos): Obsidian CLI transport never engaged; flock(1) absent breaks all vault locking

willyfog766:
Three defects that share one root cause: **the plugin was never exercised end-to-end on macOS, and CI only runs on `ubuntu-latest`.** A Linux-only matrix is structurally incapable of catching any of these. I found them by running the code on a Mac, not by reading it.

No version bump and no tag — that's your call, so the CHANGELOG entry is filed under `## [Unreleased]`.

### 1. The Obsidian CLI transport silently never engaged on macOS (`scripts/detect-transport.sh`)

Detection probed only `PATH`, but Obsidian 1.12 ships the CLI *inside* the app bundle and doesn't symlink it out. So every stock macOS install fell through to the `filesystem` floor **while reporting success** — the exact silent-downgrade the transport design exists to prevent. Separately, the version probe called `obsidian-cli --version`, which the CLI rejects outright (it takes a `version` subcommand), so it could never have succeeded even if it had found the binary.

Fixed with a candidate list (`PATH` → app bundle → bare `obsidian` last) and a capability probe bounded by a 3s timeout, hand-rolled because macOS has no `timeout(1)`. Without the bound, an Electron launcher that treats a bare `version` as a file to open would hang detection — and several skills invoke it.

### 2. `flock(1)` doesn't exist on macOS, so every advisory lock in the vault failed

(`scripts/allocate-address.sh`, `scripts/wiki-lock.sh`)

`flock(1)` is util-linux. Both scripts called it directly, so address allocation and all locking died with `flock: command not found`, taking **3 of 9 `make test` targets red on any Mac**. Python was never affected — `fcntl.flock` is a POSIX syscall.

New `scripts/lib/portable-lock.sh` takes the same `flock(2)` kernel lock on fd 9 by either route: `flock(1)` where present (Linux behavior unchanged), and `python3`'s `fcntl.flock` on the inherited fd where not. The lock lives on the open file description, so it persists after python exits and the kernel releases it on shell exit — the same mechanism `flock(1)` uses internally.

**Worth recording, since it may save you from repeating it:** I first emulated this in pure shell with an atomic `mkdir` spin plus stale-lock reaping. It had a use-after-steal bug. The *fixed* version still let two workers into the critical section about **1 run in 5** under contention. The reason is that macOS ships **bash 3.2, which has no `BASHPID`** — a subshell cannot identify itself, so every PID-based reaping scheme leaks. The kernel lock deletes the entire class: no reaping, no PID tracking, no owner tokens, no `EXIT` trap, no steal semantics.

### 3. Every recipe in `skills/wiki-cli/SKILL.md` was wrong; none of them ran

The CLI takes `key=value` options, not positional args. Beyond syntax: `vault=` takes a vault *name*, not a path; there is no `write` command (it's `create` + `overwrite`); `daily:today` doesn't exist (`daily:read` / `daily:path`); and `content=` is a shell argument with `\n` escapes, not stdin — so the filesystem Write tool is genuinely better for real pages, which the docs now say. All 18 rewritten recipes were executed against a live vault.

Two traps documented that only surface by running them: `base:views` ignores `path=` and reads whatever file is active in the GUI (it isn't scriptable), and `base:query` without `view=` resolves the *first* view rather than all. I also added recipes for the CLI's native `orphans` / `unresolved` / `deadends`, which `wiki-lint` currently reimplements by hand.

### Also

BSD `wc -l` pads output with leading spaces, breaking string compares in three test files.

### Verification

- `make test` — **10/10 green on macOS**, the first time the suite has passed on the platform. Also green on Linux via the `flock` backend.
- New `tests/test_portable_lock.sh`: mutual exclusion under 6×20 contended workers, run 8× (**960 contended acquisitions, 0 violations**), plus timeout, release-on-exit, and reclaim-after-`SIGKILL`.
- **New `macos-latest` CI job.** This is the most important change here — it's the regression guard for the whole class. The bugs above are symptoms; Linux-only CI is the cause.
- `PORTABLE_LOCK_FORCE_BACKEND` + a CI step forcing the `python3` backend, because Linux runners always have `flock` and the python path would otherwise have zero coverage anywhere. The suite asserts the forced backend was *actually* taken — an earlier version of that check silently ran `flock` while claiming to test python, which is worse than no check.


willyfog766:
Pushed a second commit (`9d91431`) fixing a **fourth bug**, found after opening this PR. It is the same silent-wrong-target class as the transport bug, in a second location, and it is arguably the worst one here.

### `vault=<name>` can silently read and write a completely different directory

The CLI resolves `vault=` against **Obsidian's own vault registry**, never against your working directory. So a present, working binary does *not* mean it can reach the vault you are standing in:

- Obsidian may not have this directory open as a vault at all, or
- **another directory may be registered under the same name** — in which case every CLI call reads and writes *that* directory, and reports success.

The second case is not hypothetical. It is how I found it: a stray copy of this repo in `~/Downloads` was registered under the name `claude-obsidian`, and it silently absorbed an entire session's CLI writes while my real working tree sat untouched. `detect-transport.sh` was reporting `preferred: cli` the whole time. Anything that honored that snapshot would have written to the wrong folder.

### Fix

`detect-transport.sh` now asks the CLI which vaults it knows (`vaults verbose`), resolves symlinks on both sides, and looks for one whose path **is** this vault root.

- No match → `cli` is **not** preferred; falls back to `filesystem` with an actionable warning.
- `available.cli.present` stays `true` (the binary is real, it just cannot reach us). The new `available.cli.vault_addressable` carries the distinction.
- The verified name is published as `available.cli.vault_name`, so consumers pass a checked `vault=` instead of guessing from the directory basename. `skills/wiki-cli/SKILL.md` documents the footgun.

Also split the timeout helper into `probe_with_timeout` (exit status only) and `capture_with_timeout` (captures stdout) — capturing `vaults` through the former silently returned nothing, since it sends stdout to `/dev/null`.

`tests/test_detect_transport.sh` is new and hermetic (mocks the CLI). It asserts `cli` **is** preferred when a vault matches, is **not** preferred when the only registered vault lives elsewhere, and that the resolved name is published. `make test` is now 11 suites, green on both macOS and Linux.

### Correction to the original PR description

Where I wrote that the 18 rewritten CLI recipes were "executed against a live vault" — that is true, but it was the *stray* vault described above, not the working tree. The recipes and the findings still stand (the argument grammar, the missing `write` command, the `base:views` / `base:query` traps are all properties of the CLI itself, not of any particular vault). But the orphan/dead-link counts I cited came from that other copy. Flagging it rather than leaving it buried.

AgriciDaniel:
Thank you. Since v2.0.0 the mutation path no longer uses `flock(1)` or the per-file `wiki-lock.sh` helper; the canonical `MutationLock` uses Python's `fcntl.flock` on a pinned vault descriptor plus an atomic `mkdir`, macOS runs in CI, and native Windows refuses writes explicitly with `UNSUPPORTED_PLATFORM` (see `docs/windows-wsl.md`). The PostToolUse auto-commit hook is also gone. Closing as fixed in v2.0.0. The lock and CI halves are on main; the vault-registration probe in your `detect-transport.sh` changes is still valuable and is tracked under #141.

PR #122, opened 2026-07-13
feat(v2.0): Live Core — internet, browser, rule packs, workflows, self-knowledge

willyfog766:
## What this adds

v1.x built a vault that compounds knowledge but was **sealed**. v2.0 opens it along five axes, each with its boundary re-derived as code and proven by execution.

| Capability | Entry point |
|---|---|
| **Always-on internet** (inbound free, egress gated) | `scripts/net-policy.py` |
| **Browser** (JS/login pages WebFetch can't see) | `scripts/detect-browser.sh` |
| **Rule packs** (1 rule → 6 agent dialects) | `scripts/render-rules.py` |
| **Workflows** (Fable fan-out) | `workflows/*.js` |
| **Self-knowledge** (source-derived, execution-verified manifest) | `scripts/core-introspect.py` |

Full guide: `docs/live-core-guide.md`.

## The security boundary

"Internet at all times" collides head-on with the v1.7.0 audit's BLOCKER B1 (data-egress consent gap). Solved by the **inbound/egress asymmetry**, not a single on/off flag:

- **Inbound** (reading the web) is on by default.
- **Egress** (sending vault content out) asks every time; secrets never leave.

Boundary properties, each asserted in CI:
- SSRF guard denies loopback / private / link-local / reserved space in **every encoding a real fetch resolves** — dotted-quad, decimal (`2130706433`), hex, octal, short form. A canonical-only guard is trivially bypassed by writing the other spelling; the cloud-metadata endpoint in decimal (`2852039166`) is the dangerous case.
- Secret payloads (`.env`, `*.pem`, `*.key`, `id_rsa*`, `.aws/`, `.ssh/`) are denied **before** the consent check, so they're refused even to an allowlisted host. Matched **case-insensitively** (APFS folds `.ENV` to `.env`) and at **any path segment** (a file inside a `.env/` or `credentials/` directory).
- A corrupt policy file **fails closed**.

## Adversarial review

A pre-commit verifier pass (per `CLAUDE.md`) and independent execution testing found **two BLOCKERs**, now fixed with regression tests:
1. SSRF bypass via encoded IP literals (decimal/hex/octal/short-form).
2. Case-sensitive secret filter leaking `.ENV` on case-folding filesystems.

Plus one HIGH (a documented-but-unimplemented `session` consent mode, removed) and two MEDIUMs (secrets inside secret-named directories; half-corrupt managed block now raises instead of silently duplicating).

## Testing

- `macos-latest` added to the CI matrix — a Linux-only pipeline structurally cannot see BSD-userland bugs. This surfaced and fixed a **pre-existing** BSD `wc(1)` padding bug that broke four concurrency assertions the instant macOS entered CI.
- Every fix carries a regression test **wired into its runner** (one was caught defined-but-uncalled — the exact `green-gate-is-not-the-feature` trap) and **negative-controlled**: reverted to confirm red, restored to confirm green.
- Full suite green on macOS **for the first time**. `core-introspect verify`: **94 reachable, 5 skipped, 0 broken**.

## Not in scope (documented in the threat model)

DNS rebinding, redirect re-checking, and prompt injection from fetched content are **out of scope** for a policy layer and called out explicitly in `docs/live-core-guide.md` rather than implied as guarantees.

No version bump or tag — that's the maintainer's call.


AgriciDaniel:
Thank you for a substantial contribution. This branch targets the shell-script era that v2.0.0 replaced (command mirrors, `allowed-tools` frontmatter, per-file locks, direct worker writes), so it cannot be rebased. Closing the PR; the encoded-IP SSRF vectors are being adopted as regression tests.

Issue #116, opened 2026-07-09
SessionStart hook uses unsupported `type: "prompt"` → startup validation error


TechTripAi:
`hooks/hooks.json` defines a `type: "prompt"` hook under the `SessionStart` event:

```json
{
  "type": "prompt",
  "prompt": "If a vault is configured for this session (check CLAUDE.md for VAULT_PATH or a wiki/ folder in the current directory), silently read wiki/hot.md to restore recent context. ..."
}
```

Per the Claude Code docs, **SessionStart supports only `command` and `mcp_tool` hooks**:

> SessionStart runs on every session, so keep these hooks fast. Only `type: "command"` and `type: "mcp_tool"` hooks are supported.
> — https://code.claude.com/docs/en/hooks

### Impact

On stricter/newer clients this surfaces a `SessionStart::startup hook` validation error at launch. It also produces confusing machine-to-machine skew: a marketplace-installed copy throws the error while a git-source copy (whose cached artifact predates the hook) stays silent — despite both reporting `v1.9.2`.

### Note — likely redundant

The invalid hook appears largely redundant. The sibling `command` hook in the same block already restores the in-vault case:

```json
{ "type": "command", "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true" }
```

The only behavior unique to the prompt hook is the **cross-project** restore (resolving a vault referenced from another project's `CLAUDE.md`), which (a) can't be a `prompt` hook on `SessionStart`, and (b) would need a `command` equivalent that matches your vault-path convention — I noticed the hook text says `VAULT_PATH` but the docs use `Path:`, so I didn't want to presume.

### Possible fixes (your call)

1. **Strip** the prompt hook — minimal; the in-vault restore is already covered by the existing command hook. Loses only the cross-project nicety.
2. **Convert** to a `command` hook that resolves the cross-project vault path per your intended convention — preserves the feature.

_(The valid `PostCompact` prompt hook is unaffected — that event does support prompt hooks.)_

AgriciDaniel:
[...]

Issue #106, opened 2026-07-02
hooks: SessionStart prompt-type hook errors on every session start; PostCompact prompt hook never fires


iVintik:
## Problem

`hooks/hooks.json` registers two `type: "prompt"` hooks that current Claude Code (v2.x) cannot execute:

1. **SessionStart** (`matcher: "startup|resume"`) — prompt-type hooks are not supported for SessionStart. Every session start in every project logs:
   ```
   Failed to run: prompt-type hooks are not supported for SessionStart events (no conversation context is available). Use a command-type hook instead.
   ```
   Observed in 33 of 34 sessions in one month of transcripts.

2. **PostCompact** — prompt-type hooks are not supported there either (`Prompt stop hooks are not yet supported outside REPL`), so the hot-cache restore after context compaction — the one moment it matters most — has never actually run.

## Suggested fix

- **Delete** the SessionStart prompt hook: it is redundant. The sibling command hook (`[ -f wiki/hot.md ] && cat wiki/hot.md || true`) already injects the hot cache as context on startup/resume.
- **Replace** the PostCompact prompt hook with a command hook on `SessionStart` with `matcher: "compact"` (supported, command-type, stdout becomes context):
  ```json
  "SessionStart": [
    { "matcher": "startup|resume|compact",
      "hooks": [ { "type": "command", "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true" } ] }
  ]
  ```
  (or keep `compact` as a separate matcher entry if you want different behavior there).

That reduces the hook surface to one command hook that works in all three cases and removes the startup error noise entirely.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rubendeheras:
Still reproducing on `claude-obsidian@1.9.2` (latest, confirmed via `claude plugin marketplace update`). Same error on every session start:

```
Failed to run: prompt-type hooks are not supported for SessionStart events (no conversation context is available). Use a command-type hook instead.
```

The sibling `command` hook (`cat wiki/hot.md`) already injects the hot cache successfully, so the `prompt`-type entry is confirmed redundant as well as broken. This is one of ~14 open duplicates of the same root cause (#97, #96, #93, #91, #88, #87, #61, #48, #47, #45, #40, #29, #7) plus two closed-but-recurring ones (#10, #42) — might be worth a pinned/canonical issue to consolidate.

AgriciDaniel:
[...]

Issue #97, opened 2026-06-17
SessionStart prompt-type hook fails: "prompt-type hooks are not supported for SessionStart events"


Hameishiv:
## Summary

The `SessionStart` hook in `hooks/hooks.json` includes a `prompt`-type hook, which current Claude Code rejects. Every session start prints:

```
Failed to run: prompt-type hooks are not supported for SessionStart events (no conversation context is available). Use a command-type hook instead.
```

## Root cause

`SessionStart` fires before any conversation context exists, so `prompt`/`agent` hooks (which require an LLM turn to evaluate) are invalid there — only `command`-type hooks are allowed. The offending entry (v1.9.2 `hooks/hooks.json`, in the `SessionStart` → `startup|resume` block):

```json
{
  "type": "prompt",
  "prompt": "If a vault is configured for this session ... silently read wiki/hot.md to restore recent context. ..."
}
```

(The `PostCompact` prompt hook is fine — `prompt` hooks *are* supported on that event.)

## Suggested fix

The `prompt` hook is effectively **redundant**: the `command` hook directly above it already injects the file via `[ -f wiki/hot.md ] && cat wiki/hot.md`, so `hot.md` is in context regardless. Simplest fix is to drop the `SessionStart` `prompt` hook entirely. If you want to keep the "only when a vault is configured" conditional behaviour, move that instruction to a `UserPromptSubmit` hook (which supports `prompt`) instead.

## Repro

1. Install the plugin.
2. Start any Claude Code session (`startup` or `resume`).
3. The error appears among the SessionStart hook results.

Environment: Claude Code on Windows 11, plugin v1.9.2.

---
_Claude Code_

AgriciDaniel:
Thank you for the report. This was the prompt-type SessionStart/PostCompact hook shipped through v1.9.2. v2.0.0 (commit 6784d9e) removed both prompt hooks; `hooks/hooks.json` now declares only command hooks (SessionStart with the `startup|resume|clear|compact` matcher, and Stop), and `tests/test_hooks.py` pins that shape. Per the Claude Code hooks documentation, SessionStart plain stdout is added to Claude's context, so no `additionalContext` output is needed.

Note that hot-cache injection at session start is opt-in: set `CLAUDE_OBSIDIAN_SESSION_CONTEXT=1`. Closing as fixed in v2.0.0; please reopen if you still see the error on a current release.

Issue #96, opened 2026-06-16
SessionStart prompt-type hook fails on every session start (unsupported event)


hypyc95:
## Summary

The plugin's `hooks/hooks.json` registers a `"type": "prompt"` hook under `SessionStart`. Claude Code does not support prompt-type hooks for `SessionStart` (there is no conversation context at session start), so it fails on every session start with:

```
SessionStart:startup hook error
Failed to run: prompt-type hooks are not supported for SessionStart events
(no conversation context is available). Use a command-type hook instead.
```

## Location

`hooks/hooks.json` → `SessionStart` → matcher `startup|resume`, the 3rd hook entry:

```json
{
  "type": "prompt",
  "prompt": "If a vault is configured for this session (check CLAUDE.md for VAULT_PATH or a wiki/ folder in the current directory), silently read wiki/hot.md to restore recent context. ..."
}
```

## Impact

Non-blocking — the two sibling `command` hooks in the same block still run — but it surfaces an error banner on every session start.

## Note

The prompt hook appears redundant: the first `command` hook in the same block already runs `[ -f wiki/hot.md ] && cat wiki/hot.md` to restore the hot cache into context.

## Suggested fix

Remove the prompt-type hook from the `SessionStart` block. If the restore-into-context behavior is needed beyond the `cat`, move it to a `SessionStart`-compatible mechanism (the command hook already injects file contents), or relocate the prompt hook to a supported event such as `UserPromptSubmit`.

## Environment

- Claude Code on Windows 11
- claude-obsidian v1.9.2

AgriciDaniel:
Thank you for the report. This was the prompt-type SessionStart/PostCompact hook shipped through v1.9.2. v2.0.0 (commit 6784d9e) removed both prompt hooks; `hooks/hooks.json` now declares only command hooks (SessionStart with the `startup|resume|clear|compact` matcher, and Stop), and `tests/test_hooks.py` pins that shape. Per the Claude Code hooks documentation, SessionStart plain stdout is added to Claude's context, so no `additionalContext` output is needed.

Note that hot-cache injection at session start is opt-in: set `CLAUDE_OBSIDIAN_SESSION_CONTEXT=1`. Closing as fixed in v2.0.0; please reopen if you still see the error on a current release.

PR #124, opened 2026-07-14
fix(hooks): remove invalid prompt-type hooks and restore post-compaction hot cache via the SessionStart compact matcher

### LINKED ISSUE #40: SessionStart prompt hook fails with "ToolUseContext is required for prompt hooks"

### LINKED ISSUE #110: [bug] Plugin fails to load in Claude Code 2.1.199: `PostCompact` is not a valid hook event
## Summary
  
  After installing `claude-obsidian@1.9.2` on Claude Code `2.1.199`, the plugin fails to load with a Zod schema error. The root cause is that `hooks/hooks.json` declares a `PostCompact` hook, but `PostCompact` is **not** a valid Claude Code hook event — the schema only accepts `PreCompact`. Because Claude Code
  validates the entire `hooks.json` as a single record, a single invalid key rejects the whole file, so **all four hooks (SessionStart, PostToolUse, Stop) also stop working**, not just the one that's malformed.

  ## Environment

  - Claude Code: `2.1.199` (latest at time of report)
  - Plugin: `claude-obsidian@1.9.2`
  - OS: macOS 14.4 (darwin 23.4.0), APFS (case-insensitive)
  - Install method: `claude plugin marketplace add` + `claude plugin install`

  ## Reproduction

  ```bash
  claude plugin marketplace add AgriciDaniel/claude-obsidian
  claude plugin install claude-obsidian@agricidaniel-claude-obsidian
  claude plugin list
  ```

  Output:

  ```
  ❯ claude-obsidian@agricidaniel-claude-obsidian
      Version: 1.9.2
      Scope: user
      Status: ✘ failed to load
      Error: Hook load failed: [
        {
          "origin": "record",
          "code": "invalid_key",
          "issues": [
            {
              "code": "invalid_union",
              "errors": [[{
                "code": "invalid_value",
                "values": [
                  "PreToolUse", "PostToolUse", "PostToolUseFailure",
                  "Notification", "UserPromptSubmit",
                  "SessionStart", "SessionEnd", "Stop",
                  "SubagentStart", "SubagentStop",
                  "PreCompact",
                  "PermissionRequest", "Setup",
                  "TeammateIdle", "TaskCompleted",
                  "Elicitation", "ElicitationResult",
                  "ConfigChange",
                  "WorktreeCreate", "WorktreeRemove",
                  "InstructionsLoaded"
                ],
                "message": "Invalid option: expected one of \"PreToolUse\"|...|\"PreCompact\"|..."
              }]]
            }
          ],
          "path": ["hooks", "PostCompact"],
          "message": "Invalid key in record"
        }
      ]
  ```
  
  ## Root cause

  `hooks/hooks.json` line 22:

  ```json
  "PostCompact": [
    {
      "matcher": "",
      "hooks": [
        {
          "type": "prompt",
          "prompt": "Hook-injected context does not survive context compaction. If wiki/hot.md exists in the current directory, silently re-read it now to restore the hot cache. Do not announce this."
        }
      ]
    }
  ]
  ```

  Per the [Claude Code hooks documentation](https://docs.claude.com/en/docs/claude-code/hooks), only `PreCompact` is a valid event — there is no `PostCompact`. The v1.4 release notes (`wiki/meta/claude-obsidian-v1.4-release-session.md`) describe the intent as *"re-inject hot.md **after** context compaction"*,
  which suggests the author may have assumed a `PostCompact` event existed by symmetry with `PreCompact`, or confused it with a hook event from another tool (e.g., an IDE plugin).

  The `.github/copilot-instructions.md` in this repo actually lists `PostCompact` as a "valid event name" (line 42), which is why the mistake wasn't caught.

  ## Impact

  Because `hooks.json` is validated as a whole record:
  - `SessionStart` (loads `wiki/hot.md` at startup) — broken
  - `PostToolUse` (auto-commit wiki edits) — broken
  - `Stop` (prompt to update `hot.md` on session end) — broken
  - `PostCompact` — never worked; was invalid from the start

  So **the plugin currently does nothing on Claude Code ≥ some version that tightened Zod validation**.

  ## Suggested fixes

  **Option 1 — Simplest**: remove the `PostCompact` block entirely. The other three hooks will start loading again. Users lose mid-session cache restoration after `/compact`, but that feature never worked anyway.

  **Option 2 — Preserve intent**: move the prompt into a `PreCompact` hook that pre-loads `wiki/hot.md` into the compaction summary, so it survives compaction as part of the summary itself. This actually works within the schema.

  **Option 3 — Request the event upstream**: file a feature request with Anthropic for a `PostCompact` event, and gate the hook behind a version check until it lands.

  I'd suggest Option 1 as an immediate patch release (v1.9.3), plus Option 2 as a follow-up if the author wants to preserve the mid-session restore behavior.

  Also worth fixing:
  - `.github/copilot-instructions.md:42` — remove `PostCompact` from the valid-events list
  - `hooks/README.md:10` — remove the `PostCompact` row
  - `wiki/meta/claude-obsidian-v1.4-release-session.md` — mark the v1.4 `PostCompact` addition as broken/reverted

  ## Local workaround

  For anyone hitting this before the fix ships:

  ```bash
  # Edit ~/.tclaude/plugins/marketplaces/agricidaniel-claude-obsidian/hooks/hooks.json
  # and delete the "PostCompact": [ ... ] block (lines 22–32).
  # Then restart Claude Code.
  ```
  
  Happy to open a PR with Option 1 if that's welcome.

j4holt:
## Problem

`hooks/hooks.json` ships two `"type": "prompt"` hook entries that Claude Code rejects:

1. **`SessionStart`** — prompt-type hooks are only valid on decision-capable events (`PreToolUse`, `PermissionRequest`, `Stop`). On `SessionStart`, Claude Code errors on every startup/resume:
   ```
   SessionStart:startup hook error — prompt-type hooks are not supported for
   SessionStart events (no conversation context is available). Use a command-type hook instead.
   ```
2. **`PostCompact`** — the prompt hook there never worked (silently dropped), and newer Claude Code versions (≥ 2.1.199, see #110) reject `PostCompact` as an invalid hook event outright, which breaks plugin load.

Already reported in #7, #29, #40, #45, #47, #48, #61, #87, #88, #91, #93, #96, #97, #106, #110, #116.

## Why this PR instead of the existing ones

The open fixes (#13, #21, #43, #49, #68, #71, #90, #99, #100) all either drop post-compaction restoration or keep a command hook on the `PostCompact` event — which still fails on Claude Code versions that reject `PostCompact` entirely (#110).

This PR uses the supported channel for the same behavior: the `SessionStart` event fires with matcher **`compact`** after context compaction, and its command-hook stdout **is** injected into context. So:

- **Removed** the `SessionStart` prompt entry — redundant; the existing command hook `[ -f wiki/hot.md ] && cat wiki/hot.md` already restores the hot cache.
- **Removed** the `PostCompact` block entirely (fixes #110's load failure).
- **Extended** the `SessionStart` matcher from `startup|resume` to `startup|resume|compact`, so the hot cache is restored after compaction — for the first time actually working, since the old prompt hook was silently dropped.
- Updated `hooks/README.md` events table to match.

Net diff: 3 insertions, 19 deletions.

Fixes #40. Fixes #110.

### DIFF
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -6,8 +6,7 @@ Plugin hooks for the claude-obsidian wiki vault. All hooks are defined in `hooks
 
 | Event | Type | Purpose |
 |---|---|---|
-| `SessionStart` | command + prompt | Loads `wiki/hot.md` into context. Command type runs `[ -f wiki/hot.md ] && cat wiki/hot.md` as the canonical safety check (works for non-vault sessions without erroring). Prompt type complements with semantic context restoration. Matcher: `startup\|resume`. |
-| `PostCompact` | prompt | Re-loads `wiki/hot.md` after context compaction. Hook-injected context does NOT survive compaction (only `CLAUDE.md` does), so this hook restores the hot cache mid-session. |
+| `SessionStart` | command | Loads `wiki/hot.md` into context via `[ -f wiki/hot.md ] && cat wiki/hot.md` (works for non-vault sessions without erroring). Matcher: `startup\|resume\|compact` — the `compact` matcher fires after context compaction, restoring the hot cache mid-session (hook-injected context does NOT survive compaction; only `CLAUDE.md` does). |
 | `PostToolUse` | command | Auto-commits any wiki/ or .raw/ changes after Write or Edit tool calls. Guarded by `[ -d .git ]` so it never errors in non-git directories, and by `git diff --cached --quiet` so it never creates empty commits. |
 | `Stop` | prompt | Updates `wiki/hot.md` at the end of every Claude response with a brief summary of what changed. |
 
@@ -15,7 +14,7 @@ Plugin hooks for the claude-obsidian wiki vault. All hooks are defined in `hooks
 
 `anthropics/claude-code#10875` documents that **plugin hook STDOUT may not be captured** by Claude Code, while identical inline hooks in `settings.json` work correctly.
 
-**Impact**: If this bug is active in your Claude Code version, the prompt-type SessionStart and PostCompact hooks may not inject context as expected.
+**Impact**: If this bug is active in your Claude Code version, the SessionStart command hooks (including the `compact`-matcher hot-cache restore) may not inject context as expected.
 
 **Workaround**: The command-type SessionStart hook (`cat wiki/hot.md`) is the canonical safety check. It relies on STDOUT capture for context injection, so test against this issue if hot cache restoration fails. As a fallback, copy the hook config from `hooks.json` into your user-level `~/.claude/settings.json` instead of relying on plugin hooks.
 
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -2,7 +2,7 @@
   "hooks": {
     "SessionStart": [
       {
-        "matcher": "startup|resume",
+        "matcher": "startup|resume|compact",
         "hooks": [
           {
             "type": "command",
@@ -11,21 +11,6 @@
           {
             "type": "command",
             "command": "[ -x scripts/wiki-lock.sh ] && bash scripts/wiki-lock.sh clear-stale --max-age 3600 >/dev/null 2>&1 || true"
-          },
-          {
-            "type": "prompt",
-            "prompt": "If a vault is configured for this session (check CLAUDE.md for VAULT_PATH or a wiki/ folder in the current directory), silently read wiki/hot.md to restore recent context. If wiki/hot.md does not exist, do nothing. This is a non-vault session. Do not announce this. Do not report what you read. Just have the context available."
-          }
-        ]
-      }
-    ],
-    "PostCompact": [
-      {
-        "matcher": "",
-        "hooks": [
-          {
-            "type": "prompt",
-            "prompt": "Hook-injected context does not survive context compaction. If wiki/hot.md exists in the current directory, silently re-read it now to restore the hot cache. Do not announce this."
           }
         ]
       }


AgriciDaniel:
Thank you for this fix. v2.0.0 (commit 6784d9e) resolved the same defect by removing the prompt-type hooks and moving the hook logic into the Python adapter, so this diff no longer applies to main. Closing as superseded by that release; the analysis here was useful in confirming the direction. The SessionStart `compact` matcher you proposed is exactly what main adopted.

Issue #125, opened 2026-07-15
[bug] SessionStart prompt-type hook errors on every session startup


mashkovtsevlx:
## What happened

Every session start emits a hook error:

```
SessionStart:startup hook error
Failed to run: prompt-type hooks are not supported for SessionStart events
(no conversation context is available). Use a command-type hook instead.
```

`hooks/hooks.json` registers three hooks under the `SessionStart` event
(matcher `startup|resume`). The third is `type: "prompt"`:

```json
{
  "type": "prompt",
  "prompt": "If a vault is configured for this session ... silently read wiki/hot.md ..."
}
```

Claude Code does not support prompt-type hooks on `SessionStart` — there is no
conversation context to inject a prompt into at startup, so the harness rejects
it and surfaces the error above. The two command-type hooks in the same block
run fine.

## Expected behavior

A clean session start with no hook error. Hot-cache restoration at startup is
already handled by the first command hook
(`[ -f wiki/hot.md ] && cat wiki/hot.md`), so the prompt hook is redundant at
this event. (Prompt-type hooks remain valid — and are correctly used — under
`PostCompact` and `Stop`.)

## Reproduction steps

1. Enable the plugin globally (`enabledPlugins` in `~/.claude/settings.json`).
2. Start any Claude Code session (`startup` or `resume`) in any directory.
3. Observe the `SessionStart:startup hook error` in the transcript.

The directory does not need to be a vault — the error fires regardless of
whether `wiki/hot.md` exists, because it's the hook *type* that's rejected, not
the hook body.

## Environment

- Plugin version: v1.9.2
- Claude Code version: (current release; behavior is version-independent — the
  harness has never supported prompt hooks at SessionStart)
- OS: macOS 15.4 (Darwin 25.4.0) — not OS-specific
- Obsidian version: n/a (fires in non-vault sessions too)
- Transport selected: n/a

## Skill / agent / script involved

`hooks/hooks.json` — the `SessionStart` block. Documented in `hooks/README.md`
(the `SessionStart | command + prompt` row).

## Logs / output

```
SessionStart:startup hook error
Failed to run: prompt-type hooks are not supported for SessionStart events
(no conversation context is available). Use a command-type hook instead.
```

## What you tried

- [x] Confirmed the two command hooks in the same block run without error
- [x] Confirmed the error fires in non-vault directories (type-level rejection)
- [x] Searched existing issues for similar reports

Note: this is distinct from the STDOUT-capture bug documented in
[...]

Issue #93, opened 2026-06-14
[bug] -  SessionStart/PostCompact prompt-type hooks fail on startup: "prompt-type hooks are not supported"


davidmjackson:
## What happened

On every session start, Claude Code prints a hook error:

SessionStart: startup hook error  Failed to run: prompt-type hooks are not supported

The plugin's hooks/hooks.json declares hooks of "type": "prompt" in its SessionStart block (one of three hooks) and as the only hook in its PostCompact block. The current Claude Code harness only executes "type": "command" hooks and rejects prompt-type ones, producing the error. As a side effect, the PostCompact block — being entirely prompt-type — is dropped completely, so wiki/hot.md is not re-read after a context compaction.

## Expected behavior

The plugin's startup and post-compact context-restoration should run without error, and the hot-cache (wiki/hot.md) should be reloaded on both SessionStart and PostCompact without printing a startup hook error.

## Reproduction steps

1. Enable claude-obsidian@agricidaniel-claude-obsidian (v1.9.2).
2. Start a new Claude Code session (claude).
3. Observe the startup banner — the SessionStart: startup hook error  Failed to run: prompt-type hooks are not supported message appears.

## Environment

- Plugin version: v1.9.2
- Claude Code version: 2.1.177
- OS: Linux (WSL2, kernel 6.6.87.2-microsoft-standard-WSL2)
- Obsidian version: n/a (error occurs independent of Obsidian; triggered in a non-vault project dir)
- Transport selected: n/a — no .vault-meta/transport.json in the working directory (non-vault session)

## Skill / agent / script involved

hooks/hooks.json — the SessionStart and PostCompact hook definitions.

## Logs / output

SessionStart: startup hook error  Failed to run: prompt-type hooks are not supported

Offending entries in hooks/hooks.json (v1.9.2):

// SessionStart — 3rd hook in the block
{ "type": "prompt", "prompt": "If a vault is configured for this session ... silently read wiki/hot.md ..." }

// PostCompact — only hook in the block
{ "type": "prompt", "prompt": "Hook-injected context does not survive context compaction. If wiki/hot.md exists ... silently re-read it now ..." }

## What you tried

- [x] Ran make test — all 8 suites green? N/A (issue is in the hook manifest, not the test surface)
- [x] Checked CHANGELOG for known issues at your version
- [x] Searched existing issues for similar reports

## Additional context

[...]

Issue #29, opened 2026-04-30
SessionStart prompt-type hook fails with 'ToolUseContext is required for prompt hooks'


nitayk:
## Summary

The plugin's `hooks/hooks.json` defines a `SessionStart` hook with two entries — a `"type": "command"` bash one-liner and a `"type": "prompt"` directive that asks Claude to silently read `wiki/hot.md`. On every session start (and on resume), Claude Code surfaces an error:

```
SessionStart:startup hook error
Failed to run: ToolUseContext is required for prompt hooks. This is a bug.
```

(and the same on `:resume`).

The error originates inside Claude Code: prompt-type hooks at SessionStart don't have a ToolUseContext available, and the hook runner refuses to dispatch them. So the bash hook runs fine, but the prompt hook always fails noisily.

## Repro

1. Install `claude-obsidian` (any 1.6.x).
2. Start Claude Code anywhere.
3. Observe the `SessionStart:startup hook error` line in the welcome banner.
4. Restart / resume — same error fires on `SessionStart:resume`.

## Affected file

`hooks/hooks.json`, the SessionStart entry's second hook:

```json
{
  "type": "prompt",
  "prompt": "If a vault is configured for this session ... silently read wiki/hot.md ..."
}
```

## Workaround I'm using locally

Strip the prompt hook, keep the bash one-liner. The plugin still loads `wiki/hot.md` on session start via the bash command — it just can't pre-prime Claude's context. Concretely:

```bash
jq '.hooks.SessionStart |= map(.hooks |= map(select(.type != "prompt")))' \
  hooks/hooks.json > hooks/hooks.json.new && mv hooks/hooks.json.new hooks/hooks.json
```

## Suggested fix

Two options that would land cleanly in the plugin:

1. **Drop the prompt hook entirely.** The bash hook already prints `wiki/hot.md` to the session, which Claude reads as initial context. The prompt-type addition is redundant for that purpose and the only thing the prompt does is reinforce silence.
2. **Move the prompt directive into `CLAUDE.md` / a SKILL** so it's part of the session's instructions rather than a hook. That's where prompt-type behavior actually works today.

Either eliminates the recurring hook error without losing functionality.

## Environment

- Claude Code: 2.1.119
- claude-obsidian: 1.6.0
- macOS 25.4.0 (Darwin)

Happy to send a PR if option (1) is the preferred direction.

Rsamus:
[...]

PR #128, opened 2026-07-16
fix(hooks): remove unsupported prompt-type SessionStart hook; convert PostCompact to command

jfcvb68nbb-max:
Fixes the `SessionStart: prompt-type hooks are not supported` startup error (tracked in #93 and ~10 duplicates: #125, #116, #106, #97, #96, #87, #61, #40, #29).

## Problem
`hooks/hooks.json` declares `type: "prompt"` hooks that the Claude Code / Codex harness rejects — prompt hooks require conversation context, which these events don't provide:
- **SessionStart** — 3rd hook is prompt-type → printed as a startup error every session.
- **PostCompact** — its only hook is prompt-type → the whole block is dropped, so `wiki/hot.md` is never re-read after a compaction (the feature silently never worked).

## Fix
- **SessionStart**: delete the prompt hook. It was redundant — the first command hook (`[ -f wiki/hot.md ] && cat wiki/hot.md`) already prints that context to stdout.
- **PostCompact**: convert the prompt hook to the same command form, so hot-cache restoration actually fires after compaction.

No behavior lost: `hot.md` still restores on both SessionStart and PostCompact, now via command hooks that print to stdout. `PostToolUse` and `Stop` hooks are untouched.

## Validation
- `hooks/hooks.json` parses as valid JSON.
- SessionStart types → `["command","command"]`; PostCompact → `["command"]`; zero prompt-type hooks remain.
- Reproduced and verified the fix on Windows 11 (Claude Code + Codex CLI); the original reporter saw it on WSL2 — platform-independent.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -11,10 +11,6 @@
           {
             "type": "command",
             "command": "[ -x scripts/wiki-lock.sh ] && bash scripts/wiki-lock.sh clear-stale --max-age 3600 >/dev/null 2>&1 || true"
-          },
-          {
-            "type": "prompt",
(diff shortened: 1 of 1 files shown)


AgriciDaniel:

PR #131, opened 2026-07-19
fix(wiki-cli): correct Obsidian CLI syntax in docs

queredda:
The Obsidian CLI takes `key=value` parameters, not positional arguments, and has no `write` command. Every documented invocation fails as written:

```bash
obsidian-cli read "$VAULT" "$NOTE"     # documented — errors
obsidian-cli read vault=MyVault path=Home.md   # actual
```

Fixed the recipe table in `wiki-cli`, plus the transport line in `autoresearch`, `save`, `wiki-ingest`, `wiki-lint`, and `wiki-query` — five skills inline the same broken form, so fixing `wiki-cli` alone would not have helped.

Also swapped `write` → `create ... overwrite`, `bases` → `base:query`, `daily:today` → `daily:path`, and noted that the CLI drives a running Obsidian instance and that `vault=` takes a vault name rather than a filesystem path.

Docs only, no behavior change. Verified against the Obsidian 1.12 CLI.


AgriciDaniel:

PR #132, opened 2026-07-19
docs(defuddle): kepano now ships a defuddle skill

queredda:
`skills/defuddle/SKILL.md` states that kepano/obsidian-skills "does not ship a defuddle skill". It does — `skills/defuddle/` is in that marketplace today.

Replaced the claim with how the two actually differ: kepano's is a short CLI reference, this one adds the vault-side workflow (`.raw/` capture, provenance header, ingest hand-off). Both can be installed.

One line, docs only.


AgriciDaniel:

PR #133, opened 2026-07-19
fix(detect-transport): probe CLI version with `version`, not `--version`

queredda:
`detect-transport.sh` probes the CLI with `obsidian-cli --version`. That flag does not exist — the subcommand is `version` — and the mistake is invisible because the binary reports errors two different ways:

```
$ obsidian-cli --version          # Obsidian running
Error: Command "--version" not found. Did you mean: version?
rc=0                              # stdout, and exit 0

$ obsidian-cli version            # no reachable Obsidian
unable to find Obsidian
rc=1
```

The first case defeats both `2>/dev/null` and `|| echo unknown`, so the error text ends up in `transport.json` as the version string:

```json
"version_string": "Error: Command \"--version\" not found. Did you mean: version?"
```

Since neither the exit code nor a message blacklist is reliable, this validates the output shape instead: accept a leading digit, otherwise `unknown`. `|| true` keeps a failing probe from aborting under `set -euo pipefail`.

After: `"version_string": "1.12.7 (installer 1.12.7)"`.

**On the `obsidian`-branch gate.** I first read `if obsidian --version >/dev/null 2>&1` as vacuous and deleted it. It is not — the invalid flag exits 0 only while Obsidian is running, so the gate is really a reachability check. It is kept, spelled with `cli_version`. Not a guaranteed no-op for every binary named `obsidian`: a GUI launcher that ignores unknown flags would pass the old gate and may fail the new one, falling back to filesystem. That is a safe false negative rather than recording a transport as usable when it is not.

The `obsidian-cli` branch stays ungated on version, as on main. That asymmetry looks like a separate design question, so I left it out of this PR.

Verified against Obsidian 1.12.7, and reviewed against an environment with no running Obsidian instance — which is how both failure modes surfaced.

### DIFF
--- a/scripts/detect-transport.sh
+++ b/scripts/detect-transport.sh
@@ -58,7 +58,28 @@ log() { $QUIET || echo "$@" >&2; }
 # json_escape: read stdin and emit a JSON-encoded string (including the
 # surrounding double quotes). Used for any untrusted value that lands in the
 # transport.json heredoc — newlines, backslashes, control chars in upstream
-# binaries (obsidian-cli --version) would otherwise break the JSON.
+# binaries (obsidian-cli version) would otherwise break the JSON.
+# cli_version: the Obsidian CLI exposes `version` as a subcommand, not a
+# `--version` flag. Two things make error detection awkward, so we validate the
+# output shape instead of trusting the exit code or blacklisting messages:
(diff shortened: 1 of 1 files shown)


AgriciDaniel:

PR #134, opened 2026-07-19
fix(defuddle): every documented invocation is missing `parse` and `--md`

queredda:
`parse` is the only command the CLI exposes, so every documented invocation fails:

```
$ defuddle https://example.com/article
error: unknown command 'https://example.com/article'
rc=0
```

It exits 0 while failing, so a `defuddle ... || fallback` guard never fires — the same reason this stayed hidden.

Second issue: `parse` emits HTML unless `--md` is passed.

```
$ defuddle parse https://example.com
<body><p>This domain is for use in documentation examples...</p></body>
```

So the recipes redirecting into `.raw/articles/*.md` were writing HTML into files named `.md`, and "Outputs clean markdown to stdout" was not accurate.

Fixed all five invocations (four in the skill, one in `wiki-ingest`) and noted the default format. The frontmatter recipe now round-trips:

```
---
source_url: https://example.com
fetched: 2026-07-20
---

This domain is for use in documentation examples without needing permission.

[Learn more](https://iana.org/domains/example)
```

Verified against defuddle-cli 0.1.0 (`npm install -g defuddle-cli`, as the skill already documents). Docs only.

### DIFF
--- a/skills/defuddle/SKILL.md
+++ b/skills/defuddle/SKILL.md
@@ -28,25 +28,25 @@ Verify: `defuddle --version`
 
 ### Clean a URL directly
 ```bash
-defuddle https://example.com/article
+defuddle parse https://example.com/article --md
 ```
-Outputs clean markdown to stdout.
+Outputs clean markdown to stdout. Without `--md` the output is HTML.
 
 ### Save to .raw/
 ```bash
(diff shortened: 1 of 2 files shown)


AgriciDaniel:

Issue #140, opened 2026-07-24
SessionStart hook fails: prompt-type hook not supported (v1.9.2)


mbursali-bs:
### Summary

The `SessionStart` hook block in `hooks/hooks.json` includes a `type: "prompt"` hook, which Claude Code rejects. Every session start (and resume) prints:

> Failed to run: prompt-type hooks are not supported for SessionStart events (no conversation context is available). Use a command-type hook instead.

`SessionStart` fires before any conversation exists, so there's no message stream for a prompt hook to inject into — Claude Code only allows `command`-type hooks there.

### Version

- claude-obsidian **v1.9.2**
- Affects both the marketplace clone and the cached install.

### Location

`hooks/hooks.json`, the third entry under `SessionStart` → `hooks`:

```json
{
  "type": "prompt",
  "prompt": "If a vault is configured for this session ... silently read wiki/hot.md to restore recent context ..."
}
```

### Impact

- Error surfaces on every startup/resume — noisy, looks like a broken install.
- Functionally harmless: the preceding command hook `[ -f wiki/hot.md ] && cat wiki/hot.md` already emits `wiki/hot.md` as session context, so the prompt hook is redundant for its stated purpose (restoring the hot cache).

### Suggested fix

Simplest: **remove the prompt hook** from `SessionStart` — the `cat wiki/hot.md` command hook already covers it.

If prompt-driven behavior is genuinely wanted at startup, move it to `UserPromptSubmit` (which does support prompt hooks) so it runs on the first user turn instead. Note the existing `PostCompact` prompt hook is fine and should stay — compaction happens mid-conversation where prompt hooks are supported.

### Repro

1. Install claude-obsidian v1.9.2.
2. Start (or resume) any Claude Code session.
3. Observe the "prompt-type hooks are not supported for SessionStart" error.

AgriciDaniel:
Thank you for the report. This was the prompt-type SessionStart/PostCompact hook shipped through v1.9.2. v2.0.0 (commit 6784d9e) removed both prompt hooks; `hooks/hooks.json` now declares only command hooks (SessionStart with the `startup|resume|clear|compact` matcher, and Stop), and `tests/test_hooks.py` pins that shape. Per the Claude Code hooks documentation, SessionStart plain stdout is added to Claude's context, so no `additionalContext` output is needed.

[...]

Issue #137, opened 2026-07-23
SessionStart hooks.json uses unsupported "type": "prompt" — startup error on every session (v1.9.2)


gerhard-cmd:
**Plugin version:** 1.9.2 (latest release; the entry is also present on current `main`)
**Claude Code version observed:** 2.1.218

**Error on every session start:**

> SessionStart:startup hook error: prompt-type hooks are not supported for SessionStart events

**Cause:** `hooks/hooks.json` defines three hooks under `SessionStart` (`matcher: "startup|resume"`). The third entry uses `"type": "prompt"`:

```json
{
  "type": "prompt",
  "prompt": "If a vault is configured for this session (check CLAUDE.md for VAULT_PATH or a wiki/ folder in the current directory), silently read wiki/hot.md to restore recent context. ..."
}
```

Per the official Claude Code hooks reference (https://code.claude.com/docs/en/hooks), SessionStart supports only `type: "command"` and `type: "mcp_tool"` hooks. Prompt-type hooks are valid on other events (the plugin's `PostCompact` prompt hook appears fine), but on SessionStart the entry is rejected and the error prints at every launch.

**Impact:** noisy startup error for plugin users; the hook itself never runs.

**Temporary local workaround:** removed the prompt-type entry from the installed plugin cache copy of `hooks/hooks.json` (`~/.claude/plugins/cache/.../hooks/hooks.json`). Not durable — a plugin update regenerates the cache and re-introduces the error.

**Suggested fix:** remove the prompt-type SessionStart entry — its intent (restoring `wiki/hot.md` context) already appears covered by the first command hook (`[ -f wiki/hot.md ] && cat wiki/hot.md || true`), whose stdout is injected as context at SessionStart. Alternatively, emit the guidance via a command hook using the documented `additionalContext` JSON output.

Happy to provide more detail if useful. Thanks for the plugin!

natea:
Confirming this is still happening on v1.9.2, hit on every SessionStart in a fresh directory. Agree with the suggested fix — the prompt-type entry duplicates what the first command hook (`cat wiki/hot.md`) already surfaces via stdout injection, so it can just be dropped.

AgriciDaniel:
[...]

Issue #91, opened 2026-06-12
SessionStart and PostCompact prompt-type hooks fail at runtime (v1.9.2)


super-ad:
## Symptom

Every Claude Code session resume in a project with `claude-obsidian` v1.9.2 installed prints:

```
SessionStart:startup hook error
Failed to run: prompt-type hooks are not supported for SessionStart events
(no conversation context is available). Use a command-type hook instead
```

PostCompact compaction events fail with the analogous message.

## Source

`hooks/hooks.json` ships two `"type": "prompt"` entries that the current Claude Code runtime rejects:

1. **`SessionStart`** — third hook entry (~line 16). The two preceding `command`-type entries (the `cat wiki/hot.md` and `wiki-lock.sh clear-stale` ones) succeed; this one fails noisily.
2. **`PostCompact`** — only hook entry (~line 27). Fails on every compaction event.

The exact JSON (eliding context):

```json
"SessionStart": [{
  "matcher": "startup|resume",
  "hooks": [
    { "type": "command", "command": "[ -f wiki/hot.md ] && cat wiki/hot.md || true" },
    { "type": "command", "command": "[ -x scripts/wiki-lock.sh ] && bash scripts/wiki-lock.sh clear-stale --max-age 3600 >/dev/null 2>&1 || true" },
    { "type": "prompt", "prompt": "If a vault is configured for this session..." }  ← fails
  ]
}],
"PostCompact": [{
  "matcher": "",
  "hooks": [
    { "type": "prompt", "prompt": "Hook-injected context does not survive context compaction..." }  ← fails
  ]
}]
```

## Repro

1. Install claude-obsidian v1.9.2 via the marketplace:
   ```
   claude plugin marketplace add AgriciDaniel/claude-obsidian
   claude plugin install claude-obsidian@agricidaniel-claude-obsidian
   ```
2. Open any project in Claude Code that has a `wiki/` directory.
3. SessionStart fires → observe the error in the terminal session log.
4. Run a long-enough conversation to trigger context compaction → observe the PostCompact error.

## User impact

- **SessionStart error**: cosmetic only when the user has a sibling `command`-type SessionStart hook that already injects `hot.md` via stdout. For users relying purely on this plugin's hook, the prompt-type failure means no hot.md context gets injected into the session header until a vault-side skill catches up.
[...]

PR #143, opened 2026-07-28
fix(hooks): remove unsupported prompt-type hook from SessionStart

Learfelix:
## Problem

Every session start prints:

```
Failed to run: prompt-type hooks are not supported for SessionStart events (no conversation context is available). Use a command-type hook instead.
```

Claude Code rejects `type: "prompt"` hooks on `SessionStart` — there is no conversation context at that point to inject a prompt into.

This is reported repeatedly and still open: #7, #29, #40, #45, #47, #48, #59, #61, #87, #88, #91, #93, #96, #97, #106, #116, #125, #137, #140.

## Fix

Remove the single `type: "prompt"` entry from `hooks.SessionStart` in `hooks/hooks.json`. Four deleted lines, nothing else.

## Why this is a no-op for functionality

The removed hook instructed the model to silently read `wiki/hot.md`. The very first hook in the same block already does exactly that, as a command hook:

```json
```

So hot-cache restore on session start is preserved. Only the error message goes away.

## What is intentionally left alone

The `PostCompact` prompt-type hook stays. `PostCompact` is still a valid hook event in Claude Code 2.1.220 (verified against the shipped binary) and it does have conversation context, so prompt-type is legitimate there. Issue #110 (`PostCompact` rejected on 2.1.199) looks like a separate regression and is out of scope for this PR.

## Verification

- `python3 -c "import json; json.load(open('hooks/hooks.json'))"` → valid
- `hooks.SessionStart` now contains only `command` hooks
- Applied the same change to a local plugin install; the startup error no longer appears, and `wiki/hot.md` is still picked up on session start
- Claude Code 2.1.220, Linux

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -11,10 +11,6 @@
           {
             "type": "command",
             "command": "[ -x scripts/wiki-lock.sh ] && bash scripts/wiki-lock.sh clear-stale --max-age 3600 >/dev/null 2>&1 || true"
-          },
-          {
-            "type": "prompt",
-            "prompt": "If a vault is configured for this session (check CLAUDE.md for VAULT_PATH or a wiki/ folder in the current directory), silently read wiki/hot.md to restore recent context. If wiki/hot.md does not exist, do nothing. This is a non-vault session. Do not announce this. Do not report what you read. Just have the context available."
           }
         ]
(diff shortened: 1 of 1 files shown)


AgriciDaniel:

PR #145, opened 2026-07-29
Fix macOS compatibility: wiki-lock.sh flock fallback + test whitespace

emircbngl:
## Problem

On macOS/BSD there is no `flock`, so `with_meta_lock` in `scripts/wiki-lock.sh` fails on the very first `acquire` with `flock: command not found`, breaking every wiki-lock operation (acquire/release/list/peek). The bundled tests also fail on macOS even when the values are correct, because BSD `wc` pads its counts with leading spaces and `assert_eq` compared them as strings (`expected '10', got '      10'`).

Repro on macOS:
```
WIKI_LOCK_VAULT=<vault> bash scripts/wiki-lock.sh acquire wiki/hot.md
# -> scripts/wiki-lock.sh: line 156: flock: command not found
```

## Changes

- **`scripts/wiki-lock.sh`** — `with_meta_lock` now detects `flock`. On Linux/util-linux the original `flock` path is unchanged. When `flock` is absent it falls back to an atomic `mkdir` lock (atomic on every POSIX filesystem), with a ~5s bounded retry matching `flock -w 5`, an age-based reap so a crashed holder can't wedge the lock, and cleanup on both normal return and an unexpected exit. A valid non-zero return (`75` = lock held) is preserved under `set -e`.
- **`tests/test_wiki_lock.sh`, `tests/test_concurrent_write.sh`** — `assert_eq` trims leading/trailing whitespace before comparing, so BSD `wc`'s padding no longer fails a correct numeric assertion.

## Verification (macOS)

- `tests/test_wiki_lock.sh`: **16/16** pass (was 15/1)
- `tests/test_concurrent_write.sh`: **6/6** pass (was 3/3)
- Concurrency invariants hold under the `mkdir` fallback: 10 concurrent writers, no leaked lockfiles, no torn/garbled lines.
- Linux behavior is unchanged (the `flock` branch is byte-identical to before).

### DIFF
--- a/scripts/wiki-lock.sh
+++ b/scripts/wiki-lock.sh
@@ -151,11 +151,34 @@ is_alive() {
 # acquire/release/clear-stale don't race against each other.
 with_meta_lock() {
   ensure_dirs
-  # Use flock under bash's redirect; meta lock is short-lived per command.
-  (
-    flock -x -w 5 9 || die "could not acquire meta-lock within 5s" 1
-    "$@"
-  ) 9>"$META_LOCK"
+  if command -v flock >/dev/null 2>&1; then
+    # Linux/util-linux: advisory fd lock, auto-released when the subshell exits.
+    (
+      flock -x -w 5 9 || die "could not acquire meta-lock within 5s" 1
+      "$@"
+    ) 9>"$META_LOCK"
+  else
+    # macOS/BSD have no flock. Portable equivalent: an atomic mkdir lock. mkdir is
+    # atomic on every POSIX filesystem, so exactly one holder wins the race. Bounded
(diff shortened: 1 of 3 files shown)


AgriciDaniel:
Thank you. Since v2.0.0 the mutation path no longer uses `flock(1)` or the per-file `wiki-lock.sh` helper; the canonical `MutationLock` uses Python's `fcntl.flock` on a pinned vault descriptor plus an atomic `mkdir`, macOS runs in CI, and native Windows refuses writes explicitly with `UNSUPPORTED_PLATFORM` (see `docs/windows-wsl.md`). The PostToolUse auto-commit hook is also gone. Closing as fixed in v2.0.0. Reaping at 5 s can delete a live holder's mutex; #144's 30 s window was safer.

PR #149, opened 2026-07-30
fix(schema): declare the page-type vocabulary once and let the router file every documented type

### LINKED ISSUE #148: Frontmatter `type` vocabulary declared 4 different ways; wiki-mode.py route rejects 6 of the 8 types WIKI.md §3 documents
## Summary

`v1.9.2` declares the frontmatter `type` vocabulary in **four** places that do not agree, and one of them is executable. The consequence is not cosmetic: `wiki-mode.py route` **rejects 6 of the 8 types** that `WIKI.md` §3 tells you to write, including two that `skills/save/SKILL.md` declares in its own template while instructing you to call that router.

An agent following the canonical docs cannot satisfy all four at once, so it picks one and the vault drifts. Mine has 126 pages of `type: session` — a value `WIKI.md` §3 does not list.

## The four declarations

| Source | Vocabulary |
|---|---|
| `WIKI.md:227` §3 | `source\|entity\|concept\|domain\|comparison\|question\|overview\|meta` |
| `skills/wiki/references/frontmatter.md:13` | same 8 (consistent with `WIKI.md`) |
| `skills/save/SKILL.md:114` | `synthesis\|concept\|source\|decision\|session` |
| `scripts/wiki-mode.py:44` `VALID_TYPES` | `source\|entity\|concept\|session\|research` |

Only the last one can reject anything — `route_path()` raises `SystemExit(4)` and the subparser uses `choices=VALID_TYPES`.

Union of all four: 12 values. Intersection: **2** (`source`, `concept`).

## Reproduction

```bash
$ cd <vault>
$ for t in session concept question comparison decision synthesis domain meta; do
    printf '%-11s exit=%s\n' "$t" "$(python3 scripts/wiki-mode.py route "$t" x >/dev/null 2>&1; echo $?)"
  done
session     exit=0
concept     exit=0
question    exit=2   # WIKI.md §3 lists it
comparison  exit=2   # WIKI.md §3 lists it
decision    exit=2   # save/SKILL.md:72 has a whole row for it
synthesis   exit=2   # save/SKILL.md:69, and :75 makes it the fallback ("When in doubt, use synthesis")
domain      exit=2   # WIKI.md §3 lists it
meta        exit=2   # WIKI.md §3 lists it
```

`skills/save/SKILL.md:35` instructs: *"consult the vault's methodology mode via `python3 scripts/wiki-mode.py route session ...`"*. For a `synthesis` note — the type that same file names as the fallback — that call exits 2 with an argparse usage dump.

## Second disagreement: the folder, for the one type that does work

`route session` is accepted, and then the two sources still disagree on where it goes:

```bash
$ python3 scripts/wiki-mode.py route session "audit closure"
wiki/sessions/audit-closure.md
```

`skills/save/SKILL.md:73` says `session` → `wiki/meta/`.

So even on the happy path the skill and the router it delegates to return different destinations, and nothing says which wins.

## Third: 11 types are written by the plugin's own templates and listed nowhere

Scanning what `skills/**` actually writes as `type:`, these appear in no enum at all:

```
area, fold, goal, moc, module, note, page, paper, project, resource, zettel
```

This looks like the root cause rather than a separate bug: the six wiki modes (§4.1a) and the four methodology modes each introduce their own vocabulary — `page` for Mode A, `module` for Mode B, `paper` for Mode E, `moc` for LYT, `area`/`resource`/`project` for PARA, `zettel` for Zettelkasten, `fold` from `wiki-fold` — and the central enum never absorbed any of them. `wiki-lint/SKILL.md:303` special-cases `type: fold` for exclusion and lines 198-199 branch on `type` not being `meta`/`fold`, so both values are load-bearing in the tooling while absent from every declared list.

## Why it matters

- **Silent drift.** No check ever fires. `wiki-lint`'s core check 6 (`skills/wiki-lint/SKILL.md:37`) verifies that `type` is *present*, never that its value is in any list — so a vault accumulates values that whichever doc you read last considers invalid. A downstream Dataview or Bases view filtering on `type` then silently misses pages.
- **Fail-closed on the documented path.** An agent that correctly follows §Mode awareness for a `synthesis` note gets `exit 2` and an argparse dump — no message explaining that the type vocabulary differs between the skill and the router.
- **Third-party linters inherit the ambiguity.** I hit this writing a schema check for my own vault: there is no answer to "is `type: session` valid?" that all four sources agree on.

## Possible direction

Not prescribing the fix, but the shape that would resolve it:

1. **One authoritative list**, with the mode-specific types folded in — most likely a base set plus per-mode extensions, since `page`/`module`/`paper`/`moc`/`area` are genuinely mode-scoped and `fold`/`meta` are structural.
2. Have `WIKI.md` §3, `skills/wiki/references/frontmatter.md`, `skills/save/SKILL.md` and `VALID_TYPES` **derive from that list** rather than restate it — four hand-maintained copies of one rule will drift again.
3. Make the router's rejection explain itself (`unknown type 'synthesis'; valid: ...`) instead of an argparse usage dump, since the docs actively point users at values it refuses.
4. Decide whether the `save` table or the router owns destination folders, and have the loser reference the winner.
5. Optional, but it is what would have caught this: a `wiki-lint` check for `type` outside the authoritative list.

Happy to send a PR for (1)+(2) if you want it in that shape.

## Environment

- Plugin `claude-obsidian` v1.9.2 (user scope, enabled)
- Claude Code on macOS 26.5 (Darwin 25.5.0), Python 3.14.5
- Vault: 1424 pages, `.vault-meta/mode.json` present with `mode: generic` (configured 2026-06-09), so the router is on its default path — not an exotic mode setup
- Related but distinct: #111 (the `scripts/` layer resolving against the plugin cache dir) touches the same layer, different cause. #70 (lint checks without a specified data-gathering method) is adjacent — a `type` check would need one.

HectorBernstorff:
Fixes #148.

## Problem

The `type` vocabulary is declared in four places that disagree. Union of 12 values, intersection of 2:

| Source | Values |
|---|---|
| `WIKI.md:70-83` | source, entity, concept, question, comparison, session, overview, meta, fold |
| `skills/wiki/references/frontmatter.md:27` | source, entity, concept, comparison, question, overview, meta |
| `skills/save/SKILL.md:42` | synthesis, concept, decision, source, session |
| `scripts/wiki-mode.py:61` `VALID_TYPES` | source, entity, concept, session, research |

The fourth is executable, so this is not only a docs inconsistency. `wiki-mode.py route` rejected **5 of the 9 types `WIKI.md` documents**, through a bare `SystemExit(4)` and an argparse usage dump:

```
route session     exit=0     route comparison  exit=2
route concept     exit=0     route overview    exit=2
route question    exit=2     route meta        exit=2
                             route fold        exit=2
```

`question` mattered most: `WIKI.md`'s root layout already ships `wiki/questions/`, and it is the most common kind of note `save` produces. A caller also could not tell a typo from a valid page type that has no filing destination — both exited the same way with no message.

## Behavior

`claude_obsidian/page_schema.py` holds the one declaration and separates two lists that behavior already distinguished but nothing wrote down:

- **`PAGE_TYPES`** — every value allowed in `type:` frontmatter.
- **`ROUTABLE_TYPES`** — the subset the router can place.

`overview`, `meta`, `fold` and `comparison` are valid frontmatter and deliberately **not** routable. That is not an omission: `transaction.py` already treats `wiki/overview.md` as one fixed path and confines a fold to `wiki/folds/` inside an `operation_type: fold` bundle. Routing them would invent destinations the rest of the product does not use. Previously `VALID_TYPES` read like the vocabulary while being neither list.

Changes:

- `wiki-mode.py` derives its accepted types from the module instead of keeping a fifth hand-maintained copy.
- `question` is routable in all four modes; `generic` gains `questions_folder`, defaulting to `wiki/questions/`.
- Rejections explain which case applies: `unknown type 'garbage'; valid page types are …` vs `type 'overview' is a valid page type but is not routable: it is created by a dedicated operation, not filed as a new note`.
- The three docs point at the module rather than restating lists. `frontmatter.md` had dropped `session` and `fold`; `save/SKILL.md` named `synthesis`/`decision`, which no other source declares.

One deliberate consequence: argparse `choices` is gone from the `route` positional, because it rejected *before* `route_path` could explain anything. `route_path` validates instead. Containment is unchanged — `route_rejection_reason` gates before `slugify`/`safe_name`/`mapping[…]`/`validate_wiki_route`, and the mapping keyspace stays provably closed (`ROUTABLE_TYPES ∪ LEGACY_TYPE_ALIASES` equals the `generic` and `para` mapping dicts' key sets). Probed with `../../../etc/passwd` as the name (→ `wiki/questions/etc-passwd.md`), and with `--evil` and `question/../../evil` as the type (both rejected before touching a path).

## Compatibility

- **`research`** stays accepted and keeps its exact prior destinations (`wiki/concepts/<name>.md` under generic, `wiki/resources/<stem>/<stem>.md` under para). It is recorded as a legacy alias but the router does **not** resolve it through the alias map, because doing so would silently relocate para-mode research folders. Asserted on the concrete paths, checked against `git show HEAD:scripts/wiki-mode.py`.
- **Existing `mode.json` files** lack `questions_folder`. Both callers deep-copy their own defaults and merge the on-disk file onto them, so the key is present by the time validation runs. The test exercises the real path — a legacy file on disk, through both `wiki-mode.py route` and `claude-obsidian.py mode get` — rather than deleting a key from an in-memory default, which is a state neither caller can reach.
- No user file is migrated, rewritten, or moved. **Rollback is reverting the commit.**

## Verification

```
make test                      # exit 0 — python + shell + contracts + package
python3 tests/test_wiki_mode.py # 31 assertions incl. 6 new
```

Regression coverage fails against the old behavior (CONTRIBUTING §1), proven by reverting the production files in an isolated copy and rerunning the new tests unmodified: `route_path('generic', 'question', …)` → `SystemExit(4)`, same for `comparison`.

Also ran `agents/verifier.md` in a fresh read-only context over the worktree. It returned `SHIP` with no BLOCKER or HIGH, and four lower findings that are all fixed in this branch rather than deferred:

- an `OPTIONAL_MODE_FOLDER_KEYS` mechanism whose stated justification it could not reproduce — removed, the key is required, and the test now exercises the caller path instead of a synthetic one;
- a `resolve_type_alias` helper with no caller — removed;
- a `PAGE_STATUSES` list with no consumer and no test, sitting under a doc sentence claiming statuses were "declared once" — removed, and the claim narrowed to types, which is what this change actually delivers;
- an inaccurate provenance comment about `mature`.

## Out of scope, noted

- **`lint_engine.py:48`** validates that `type` and `status` are *present*, never that a value means anything — which is why this divergence went undetected. A check against `PAGE_TYPES` is the natural follow-up and is left out to keep this reviewable.
- **`claude_obsidian/cli.py:546 DEFAULT_MODE_CONFIG` and `scripts/wiki-mode.py:74 DEFAULT_CONFIG` are two independent copies of the same default document** — the same "one rule, N copies" pattern this PR fixes for types, in another dimension. It predates this change (all four other generic folder keys were already duplicated), and I added `questions_folder` to both rather than unifying them, to keep the diff on one thesis. Happy to send that separately if you want it.
- **Statuses** diverge too (`WIKI.md` lists 9, `frontmatter.md` lists 4 and includes `mature`, which `WIKI.md` omits). Not touched here: picking the union or the intersection is a product decision, not a mechanical one.

## Limitations

`PAGE_TYPES` mirrors `WIKI.md`'s table, and the anti-drift test asserts every declared type appears in that document. It cannot assert the reverse — that the doc adds no undeclared type — without parsing prose, so a new row added to the table alone would not fail. If you would rather the table be generated from the module, say so and I will do that instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@ implementation record for older releases.
   or `wiki-lint.json`; CLI and vault-config patterns are combined. The
   report's new `summary.excluded_paths` count reports how many walked files
   were dropped.
+- `claude_obsidian/page_schema.py` declares the page-type vocabulary once,
+  separating every valid `type` value from the subset the methodology router can
+  file. `wiki-mode.py` derives its accepted types from it instead of
+  keeping a fifth hand-maintained copy.
+- `question` is routable in all four modes, with a `questions_folder` generic
+  setting that defaults to `wiki/questions/`.
 
 ### Changed
 
@@ -44,6 +50,13 @@ implementation record for older releases.
 
 ### Fixed
 
+- `wiki-mode.py route` rejected `question`, `comparison`, `overview`, `meta`, and
+  `fold` — five of the nine page types WIKI.md documents — and did so through a
+  bare exit with no message, so a typo and a valid-but-unroutable type were
+  indistinguishable. Rejections now explain which case applies.
+- The frontmatter reference no longer restates a shorter type list that omitted
+  `session` and `fold`, and the save skill no longer names `synthesis`/`decision`
+  types that no other source declares.
 - The scaffolded vault's `.obsidian/app.json` now pins Obsidian's "New link
   format" setting to `absolute`. It was previously left unset, defaulting to
   Obsidian's own "shortest path when possible" — links created through
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,13 @@ implementation record for older releases.
   file. `wiki-mode.py` derives its accepted types from it instead of
   keeping a fifth hand-maintained copy.
 - `question` is routable in all four modes, with a `questions_folder` generic
-  setting that defaults to `wiki/questions/`.
+  setting that defaults to `wiki/questions/`. Note the asymmetry: only `generic`
+  routes it through a configurable folder key; `para` places questions under
+  `resources_folder + "questions/"`, matching how that mode treats its siblings.
+- `wiki-mode.py route` now exits `6` for a valid-but-unroutable page type and
+  keeps `4` for an unknown one. The two are different problems with different
+  fixes, so a script branching on `$?` can tell them apart — not only a human
+  reading stderr.
 
 ### Changed
 
--- a/WIKI.md
+++ b/WIKI.md
@@ -81,6 +81,15 @@ Common page types:
 | `meta` | Index, log, cache, convention, or maintenance page |
 | `fold` | Extractive rollup of identified log entries |
 
+`claude_obsidian/page_schema.py` is the one declaration of this vocabulary. The
+table above documents it for a reader; code reads the module rather than
+restating the values, and a test asserts the two stay in step.
+
+`source`, `entity`, `concept`, `question`, and `session` are *routable*: the
+methodology router can place a new page of that type. `comparison`, `overview`,
+`meta`, and `fold` are valid frontmatter and not routable, because the product
+creates them through a dedicated operation instead of filing a new note.
+
 Common statuses are `seed`, `active`, `developing`, `evergreen`, `answered`,
 `provisional`, `contested`, `deprecated`, and `archived`. Use only statuses the
 vault's dashboards and conventions understand.
--- a/claude_obsidian/cli.py
+++ b/claude_obsidian/cli.py
@@ -566,6 +566,7 @@ DEFAULT_MODE_CONFIG: dict[str, Any] = {
             "entities_folder": "wiki/entities/",
             "concepts_folder": "wiki/concepts/",
             "sessions_folder": "wiki/sessions/",
+            "questions_folder": "wiki/questions/",
         },
     },
 }
--- a/claude_obsidian/mode_config.py
+++ b/claude_obsidian/mode_config.py
@@ -13,12 +13,12 @@ MODE_FOLDER_KEYS = {
         "entities_folder",
         "concepts_folder",
         "sessions_folder",
+        "questions_folder",
     ),
     "lyt": ("moc_folder", "notes_folder"),
     "para": ("projects_folder", "areas_folder", "resources_folder", "archives_folder"),
     "zettelkasten": ("root_folder",),
 }
-
 
 def _has_control(value: str) -> bool:
     return any(unicodedata.category(character).startswith("C") for character in value)
--- a/claude_obsidian/page_schema.py
+++ b/claude_obsidian/page_schema.py
@@ -28,6 +28,8 @@ so `VALID_TYPES` read like the vocabulary while being neither list.
 
 from __future__ import annotations
 
+from typing import NamedTuple
+
 
 # Authoritative page vocabulary. WIKI.md's table is the source this mirrors; the
 # doc now points here instead of restating the values.
@@ -43,14 +45,24 @@ PAGE_TYPES: tuple[str, ...] = (
     "fold",
 )
 
-# Types the router can place. See the module docstring for why the remainder are
-# excluded rather than missing.
-ROUTABLE_TYPES: tuple[str, ...] = (
-    "source",
-    "entity",
-    "concept",
-    "question",
-    "session",
+# The types the router deliberately does NOT place. See the module docstring for
+# why each is excluded rather than missing.
+NON_ROUTABLE_TYPES: tuple[str, ...] = (
+    "comparison",
+    "overview",
+    "meta",
+    "fold",
+)
+
+# Derived, never restated — this module exists because two copies of a list drift,
+# and writing ROUTABLE_TYPES out by hand would have reintroduced exactly that one
+# level up: nothing stopped a value from being routable without being a valid page
+# type, in which case `route_rejection` would clear a `type:` the frontmatter
+# vocabulary rejects. Subtraction makes ROUTABLE_TYPES ⊆ PAGE_TYPES structurally
+# true; a test covers the remaining direction (a typo in NON_ROUTABLE_TYPES would
+# silently leave a type routable).
+ROUTABLE_TYPES: tuple[str, ...] = tuple(
+    page_type for page_type in PAGE_TYPES if page_type not in NON_ROUTABLE_TYPES
 )
 
 # `research` was accepted by the router and documented nowhere. Its routing was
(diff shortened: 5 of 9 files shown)


HectorBernstorff:
Self-review turned up three things in my own diff. Pushed as `54a1abf` — no
force-push, no amend, so the original commit stays reviewable.

**1. The PR reintroduced its own defect, one level up.** `PAGE_TYPES` and
`ROUTABLE_TYPES` were two hand-written tuples with nothing tying them together.
A routable type that was not a valid page type would have been cleared by
`route_rejection` while the frontmatter vocabulary rejected it — two copies of a
list, drifting, which is precisely what this module exists to end.

`ROUTABLE_TYPES` is now subtraction:

```python
NON_ROUTABLE_TYPES = ("comparison", "overview", "meta", "fold")
ROUTABLE_TYPES = tuple(t for t in PAGE_TYPES if t not in NON_ROUTABLE_TYPES)
```

⊆ now holds structurally rather than by assertion. A test covers the direction
subtraction cannot: a typo in `NON_ROUTABLE_TYPES` would silently leave a type
routable and nothing else would notice.

**2. The anti-drift test was satisfied by this PR's own prose.**
`test_page_vocabulary_matches_the_documented_table` searched all of `WIKI.md`
for ``f"`{page_type}`"`` — and the same commit added a paragraph naming five of
the types in prose. **Deleting the entire table would have left it green.** It
now parses the `| Type | Purpose |` table and compares both directions, so
`WIKI.md` documenting a tenth type fails as well.

**3. The exit-code fix stopped halfway.** The stated argument is that a typo and
a valid-but-unroutable type are *different problems with different fixes* — but
both still exited `4`, so a script branching on `$?` gained nothing and only a
human reading stderr could tell them apart. `route_rejection` now returns the
code alongside the message: **`6` for unroutable, `4` unchanged for unknown**
(`2`, `3`, `4`, `5` are already in use in `wiki-mode.py`). Asserted through the
CLI, not only through the function.

Verified by mutation rather than by a green run — deleting one `WIKI.md` table
row, a ghost value in `NON_ROUTABLE_TYPES`, and collapsing the two exit codes
each fail the intended test. `make test` green before and after.

Also documented in the CHANGELOG: `questions_folder` is configurable under
`generic` only; `para` places questions under `resources_folder + "questions/"`.
That asymmetry was undocumented before.

The two follow-ups from the original description (a `type`-value check in
`lint_engine`, and unifying the duplicated `DEFAULT_*_CONFIG` pair) are still
offered and still not in this branch — happy to send either if you want them
here rather than separately.

AgriciDaniel:
Maintainer note: I rebased this branch onto current main (after PR #186) and resolved the CHANGELOG conflict by folding your Added and Fixed bullets into the existing Unreleased sections. No code changes were made. The full hermetic suite passes on the rebased branch; merging once CI is green. Thank you for the careful consolidation and the legacy `mode.json` migration test.

PR #158, opened 2026-08-04
feat(hooks): warn on Stop when wiki/log.md nears the fold batch size

twannooitmeer:
## Goal

Surface a growing, un-folded `wiki/log.md` backlog before it gets forgotten
between sessions, without ever auto-triggering a fold, since
`skills/wiki-fold/SKILL.md` is explicit that folding stays human-invoked.

## Summary

- Adds one advisory line to the Stop hook's existing bounded `systemMessage`
  once the number of `wiki/log.md` entries above the most recent `fold` marker
  reaches 85% of the wiki-fold batch size (`2**k`, default `k=4`, so 14 of 16;
  override with `CLAUDE_OBSIDIAN_FOLD_BATCH_EXPONENT`).
- Advisory only: never runs wiki-fold, never writes, never sets
  `recommend_recover`.
- Rebased onto v2.2.0. Built on #177's `recommend_recover`, so
  `_bounded_status` and every journal warning string are unchanged from main.
  No `hooks/hooks.json` changes.

## Why the Stop contract widens, and how narrowly

This moves the documented Stop behavior (`hooks/README.md`) from "recovery
state" to "recovery state plus one advisory". The reasoning:

- wiki-fold refuses automatic triggering, so a nudge is the only mechanism
  available, and Stop is the existing end-of-work status surface. SessionStart
  is opt-in egress into model context, so it is the wrong place.
- It reuses the same channel with the same bounds, adds no new output shape,
  and stays silent below the threshold.
- The line carries a count and the batch size only: no paths, no note content,
  no entry titles (asserted in tests).
- It is appended after all recovery warnings, so it can never push a recovery
  warning out of the `MAX_STATUS_ITEMS` cap, and it is skipped entirely on the
  early-return paths for unsafe metadata.

If an opt-in default (like `CLAUDE_OBSIDIAN_SESSION_CONTEXT=1`) is preferred,
that is a small follow-up.

## Cost inside the 5 second hook budget

One `_bounded_regular_bytes` read (same symlink and regular-file guard as the
hot cache and journals), capped at 512 KiB, and a bytes regex that stops at the
first marker. No decode, no second file. Worst case is bounded by the cap,
whatever the log size.

Measured against a real vault (1.1 MB `wiki/log.md`, 520 entries, Apple
Silicon, 200 runs):

| | median | max |
|---|---|---|
| `stop_status` without the check | 0.38 ms | 0.53 ms |
| `stop_status` with the check | 2.09 ms | 3.98 ms |
| full `claude-obsidian.py hook stop` process | 22 ms | 25 ms |

## How it works

wiki-fold rolls up the newest `2**k` entries and prepends its `fold |` marker
above them without trimming anything. So the entries above the topmost marker
are exactly the ones logged since the last fold. The count stops there; a vault
that was never folded counts every entry. The `fold` op token is now documented
as reserved in `skills/wiki-fold/SKILL.md`.

## Verification

- `make test` on Python 3.13: all hermetic tests and executable contracts
  passed.
- `python3 tests/test_hooks.py`: 12 new tests, covering the threshold edge
  (13 vs 14 of 16), silence right after a fold, CRLF, never-folded logs, the
  512 KiB bound, a symlinked log being ignored, no path or content in the
  output, the exponent override and invalid values, ordering after a recovery
  warning, and no `transaction recover` advice for a lock plus a backlog.
- `git diff --check` clean.

## Limitations

- The count only ever errs low: a backlog larger than the 512 KiB window, or an
  explicit-range fold of older entries, can undercount. It never overcounts.

## Rollback

Single commit; revert it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ implementation record for older releases.
 
 ## [Unreleased]
 
+### Added
+
+- Stop hook: an advisory-only line when `wiki/log.md` holds at least 85% of a
+  wiki-fold batch (`2**k` entries, default `k=4`, override with
+  `CLAUDE_OBSIDIAN_FOLD_BATCH_EXPONENT`) above its most recent `fold` marker.
+  It reads at most the first 512 KiB of the log, is appended after any
+  recovery warnings, never recommends `transaction recover`, and never runs
+  wiki-fold or writes to the vault, since folding stays human-invoked per
+  `skills/wiki-fold/SKILL.md`.
+
 ## [2.2.0] - 2026-09-10
 
 Backlog triage: lint scoping, lock recovery, host validation, the `bin/`
--- a/claude_obsidian/hook_adapter.py
+++ b/claude_obsidian/hook_adapter.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import math
 import os
 import re
 import stat
@@ -27,9 +28,13 @@ MAX_CONTEXT_BYTES = 32 * 1024
 MAX_STATUS_BYTES = 4 * 1024
 MAX_STATUS_ITEMS = 8
 MAX_TRANSACTION_SCAN = 256
+MAX_LOG_SCAN_BYTES = 512 * 1024
+DEFAULT_FOLD_BATCH_EXPONENT = 4
+FOLD_BACKLOG_WARN_FRACTION = 0.85
 OPEN_TAG = '<claude-obsidian-context trust="local-data" instructions="never">'
 CLOSE_TAG = "</claude-obsidian-context>"
 _SAFE_OPERATION_ID = re.compile(r"^[A-Za-z0-9_.-]{1,128}$")
+_LOG_ENTRY_OP = re.compile(rb"^## \[\d{4}-\d{2}-\d{2}\] (\S+)", re.MULTILINE)
 
 
 def _bounded_regular_bytes(root: Path, path: Path, limit: int) -> bytes | None:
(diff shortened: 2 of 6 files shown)


AgriciDaniel:
Thank you for this. The Stop-hook path changed on main: PR #177 refactored `_bounded_status` to take `recommend_recover` and rewrote the journal warning strings, so this diff no longer applies. Two things before it can be reviewed on its merits: please rebase onto current main, and note that this widens the documented Stop contract from recovery state to advisory nudges (`hooks/README.md`) and adds a per-Stop read of `wiki/log.md` inside the 5 second hook budget, so a short justification for both in the PR body would help. Leaving open.

twannooitmeer:
Thanks for taking a look. Rebased onto current main (v2.2.0) and updated the PR body.

**Rebase.** I dropped my own `needs_recovery` plumbing in favour of #177's `recommend_recover`. The fold check never sets it, so `_bounded_status` and all the journal warning strings are now identical to main. The `hook_adapter.py` change is additive: three constants, two small helpers, and two call sites in `stop_status`.

**A bug I found while rebasing.** My original count skipped a leading run of `fold` markers and then counted the entries below them. But wiki-fold prepends its marker *above* the entries it just folded, so that version fired straight after every fold. I replayed my own vault's log as it stood right after its last fold: the old code reported 168 unfolded entries, the fixed code reports 0. The backlog is now just the entries above the topmost marker, and there's a regression test for it.

**On your two points** (both are also in the PR body now):

- *Stop contract.* Yes, it widens the Stop row from recovery state to recovery state plus one advisory. I've kept it as narrow as I could. It uses the same channel and bounds, and the message is just a count and the batch size, with no paths or note content. It's appended after the recovery warnings, so it can't push one out of the 8-item cap. It never recommends `transaction recover`, and it's silent below the threshold. It lives in Stop because wiki-fold rules out automatic triggering, so a nudge is the only option, and Stop is where end-of-work status already goes. If you'd rather it default to off behind an opt-in variable, like `CLAUDE_OBSIDIAN_SESSION_CONTEXT`, that's a small change and I'm happy to make it.
- *5 second budget.* It's one `_bounded_regular_bytes` read capped at 512 KiB, plus a bytes regex that stops at the first marker. There's no decode and no second file. On a real 1.1 MB, 520-entry log, `stop_status` goes from a 0.38 ms median to 2.09 ms (4 ms max over 200 runs). The whole `hook stop` process takes about 22 ms, and most of that is interpreter start-up. The 512 KiB cap bounds the worst case no matter how big the log gets.

`make test` passes on Python 3.13 and `git diff --check` is clean.

PR #159, opened 2026-08-06
fix(lint): drop gitignored candidates from ambiguous wikilink resolution

Packetslave:
# Pull request

## Summary
When a vault is adopted at a repository root, gitignored build artifacts can shadow a page's basename and turn every wikilink to that page into a false `ambiguous_targets` finding (plus `stale_index_entries` from index pages), which blocks `checkpoint` with `LINT_FAILED` whenever the artifact exists. Real-world trigger: a compiled Go binary `src/env8oy/bin/env8oy` shadowing `wiki/projects/env8oy.md` made all 17 `[[env8oy]]` links ambiguous. This PR makes lint link resolution gitignore-aware: when a link resolves to multiple candidates and at least one is **not** gitignored, gitignored candidates are dropped before ambiguity is reported.

## Type
- [x] Bug fix (`fix:`)

## Related issue
None filed — happy to open one if you'd like the discussion tracked separately.

## Changes
- `claude_obsidian/gitignore.py` — **new** pure-Python `.gitignore` evaluator (`GitignoreMatcher`): documented gitignore semantics (anchoring, `!` negation, dir-only trailing `/`, `*`/`?`/`[...]`/`**`, nested-scope precedence with last-match-wins, excluded-directory containment). Reads only `.gitignore` files inside the scan root — never `.git/info/exclude`, global excludes, or a `git` subprocess — so the engine stays process-free and deterministic. Case-sensitive by design (git's default; reading `core.ignorecase` would need git config). Symlinked `.gitignore` files skipped; 1 MiB per-file cap; lazy + cached.
- `claude_obsidian/lint_engine.py` — filters gitignored candidates at the single `resolver.resolve()` call site, only when >1 candidates and a non-ignored candidate exists; module docstring updated. Report schema unchanged (`REPORT_VERSION` stays 1); no new keys.
- `tests/test_gitignore.py` — **new**, 12 matcher unit tests.
- `tests/test_lint_engine.py` — new `GitignoreDisambiguationTests` (5 tests), including a control proving the same shadow **without** a `.gitignore` still reports ambiguity, and the two easy-to-get-wrong edges: ambiguity among only-gitignored candidates is still reported, and a link whose only candidate is gitignored still resolves (no new dead links).
- `config/capabilities.json` — declares the new read surface (`vault:**/.gitignore`) in the wiki-lint capability's `read_scope`.
- `skills/wiki-lint/SKILL.md` — one sentence documenting gitignore-aware resolution.
- `CHANGELOG.md` — entry under `## [Unreleased]` / Fixed.

## Compatibility / rollback
- No report-schema change; no new findings categories. The only behavior delta: multi-candidate resolutions that previously reported ambiguity now resolve to the non-gitignored candidate(s). Vaults without `.gitignore` files are byte-identical (verified by the existing determinism tests).
- Gitignored files remain valid link targets when they are the only candidate, so no links become newly dead.
- Rollback is a clean revert of one commit; `gitignore.py` has no other consumers.

## Safety self-review
- [x] New identifiers named for the next reader
- [x] Smallest unit that works (no speculative abstraction)
- [x] Deletions kept up with additions where applicable
- [x] New behavior has hermetic test coverage
- [x] New failure modes have explicit handling + undo plan (unreadable/oversized/symlinked `.gitignore` → no rules; invalid pattern lines skipped)
- [x] Product code and mutable user-vault state remain separate
- [x] Knowledge writes use one recoverable transaction; workers only draft (n/a — lint stays read-only)
- [x] Network/destructive/external actions have explicit consent (n/a — no egress, no subprocess)
- [x] Claims and capability maturity are evidence-backed

## Testing
```
make test
```
```
All hermetic tests and executable contracts passed.
```
Run from an isolated copy of the tree (a dev checkout nested inside a user vault trips `test_vault_root_separation.py`'s cwd-discovery test on unmodified `main` too — pre-existing location artifact, unrelated to this change). Also verified against the real vault that triggered this: 16 `ambiguous_targets` + 1 `stale_index_entries` → 0/0, and `checkpoint` unblocks.

## Verifier
`agents/verifier.md` dispatched on the worktree scope:

- Verdict: SHIP
- BLOCKER: 0 / HIGH: 0 / MEDIUM: 1 / LOW: 2

The MEDIUM (undeclared `.gitignore` read surface in `config/capabilities.json`) and first LOW (document case-sensitivity in the module docstring) are fixed in this PR. The second LOW noted that a leading `^` in a character class acts as negation; git's own wildmatch also accepts `^` as class negation (`NEGATE_CLASS2`), so behavior matches git and was left as is. The verifier also independently confirmed the regression tests fail against unmodified `main` (not tautological).

## CHANGELOG
- [x] Added an entry under `## [Unreleased]` in `CHANGELOG.md`

## Screenshots / output
Before (vault with a gitignored `bin/env8oy` next to `wiki/projects/env8oy.md`):
```
ERR LINT_FAILED: vault lint has blocking findings: ambiguous_targets=16, stale_index_entries=1
```
After: lint reports 0 `ambiguous_targets`, 0 `stale_index_entries`; `checkpoint` proceeds.

## Notes for reviewer
- The filter sits at the one `resolver.resolve()` call site rather than inside `_Resolver`, so exact-match resolution, backlink (`incoming`) tracking, and fragment validation are untouched; candidate ordering still comes from `_Resolver._deduplicate`.
- Known scope limits (documented in the module docstring): only `.gitignore` files inside the vault root are read; `.git/info/exclude`, global excludes, and `core.ignorecase` are intentionally out of scope to preserve the process-free/deterministic contract.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,17 @@ implementation record for older releases.
   and accepting the content only if the bytes are identical to the first
   read. Any size, inode, device, or mode change, and any content change
   during the read, still fails closed with `CORRUPT_RUNTIME_STATE`.
+- Lint link resolution no longer reports a wikilink as ambiguous when the
+  extra candidates are gitignored files (for example a compiled binary whose
+  name shadows a page, such as `bin/env8oy` next to `wiki/projects/env8oy.md`).
+  A new pure-Python `.gitignore` evaluator (`claude_obsidian/gitignore.py`)
+  reads only `.gitignore` files inside the vault root — never
+  `.git/info/exclude`, global excludes, or a `git` subprocess — keeping
+  reports deterministic and process-free. Gitignored files remain valid link
+  targets when they are the only candidate, so no new dead links are
+  introduced; ambiguity among only-gitignored candidates is still reported.
+  This unblocks `checkpoint` runs that previously failed `LINT_FAILED`
+  whenever a build artifact existed.
 
 ## [2.1.1] - 2026-08-26
 
--- a/claude_obsidian/gitignore.py
+++ b/claude_obsidian/gitignore.py
@@ -1,0 +1,203 @@
+"""Pure-Python ``.gitignore`` evaluation for the deterministic lint engine.
+
+This module reads ``.gitignore`` files found inside a scan root and answers
+whether a relative path is ignored, following the documented gitignore
+semantics that matter for link resolution:
+
+- later patterns override earlier ones; deeper ``.gitignore`` files override
+  shallower ones;
+- ``!`` negation re-includes a previously excluded path;
+- a pattern containing a non-trailing ``/`` is anchored to its
+  ``.gitignore`` directory, otherwise it matches at any depth;
+- a trailing ``/`` restricts a pattern to directories;
+- ``*`` and ``?`` never cross ``/``; ``**`` does;
+- everything inside an excluded directory stays excluded (``.gitignore``
+  files inside excluded directories are not read).
+
+Scope is deliberately narrow and deterministic: only ``.gitignore`` files
+within the scan root are consulted — never ``.git/info/exclude``, the global
+``core.excludesFile``, or a ``git`` subprocess. Matching is always
+case-sensitive (git's default), even on case-insensitive filesystems where
+git with ``core.ignorecase=true`` would fold case; reading git config would
+break the process-free contract. Symlinked ``.gitignore`` files are skipped,
+matching the engine's symlink hygiene.
+"""
+
+from __future__ import annotations
+
+import re
+from dataclasses import dataclass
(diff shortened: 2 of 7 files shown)


AgriciDaniel:
Maintainer note: I rebased this branch onto current main (after PR #186, which added a default dot-directory skip and `--exclude` globs to the linter) and reconciled the three overlapping files so both features compose: your gitignore-aware dedup runs inside the link-resolution loop after the new exclusion filter, both notes stay in the wiki-lint skill and the engine docstring, and your changelog bullet sits under the current Unreleased Fixed list. No behavior in your commit changed and the commit remains yours. Full suite passes on the rebased branch; merging once CI is green. Thank you for the careful matcher and tests.

PR #161, opened 2026-08-08
feat: hot-sync + detach from upstream

saixso:
## Summary
- Add deterministic `scripts/hot-sync.sh` (+ skill/command/tests) so `wiki/hot.md` regenerates from git/plugin truth instead of soft LLM Stop prompts
- Wire SessionStart/Stop in Claude + Cursor hooks to run hot-sync; preserve Active Threads markers; skip write when unchanged
- Document vault-os as an independent product that no longer routinely merges `upstream/main`

## Test plan
- [ ] `make test-hot-sync` passes
- [ ] `bash scripts/hot-sync.sh` rewrites hot.md to current version/skill count
- [ ] Second run prints `unchanged` without dirtying the tree
- [ ] Active Threads bullets survive a `--force` rewrite
- [ ] Fresh Cursor/Claude session in this repo injects a truthful hot cache


saixso:
Opened by mistake against upstream; vault-os is no longer submitting PRs here.

PR #172, opened 2026-08-25
fix: stop_status misreports large complete journals as unreadable

### LINKED ISSUE #164: stop_status reports completed journals over 64 KiB as unreadable, and the recommended `transaction recover` cannot clear it
## Summary

`stop_status()` reports healthy, completed transaction journals as "unreadable" whenever the journal exceeds 64 KiB. The journals are valid JSON with `"state": "complete"` — they are simply larger than the probe's read cap, so the truncated slice fails to parse.

The warning cannot be cleared by the command it recommends, and the count grows permanently as a vault accumulates large transactions.

## Root cause

`claude_obsidian/hook_adapter.py:270` reads a bounded prefix and immediately parses it as complete JSON:

```python
payload = _bounded_regular_bytes(root, journal, 64 * 1024)
...
state = strict_json_loads(payload.decode("utf-8")).get("state")
except (UnicodeDecodeError, json.JSONDecodeError, ValueError):
    unreadable_journals += 1
```

`_bounded_regular_bytes` returns `handle.read(limit + 1)` (line 98), so any journal over 64 KiB yields a truncated 65537-byte slice, `strict_json_loads` raises `JSONDecodeError`, and the file is counted as unreadable.

The probe only needs the `state` key, but it pays the cost of parsing the entire document.

## Why 64 KiB is too small

The same package explicitly permits transactions far larger than this cap allows.

`claude_obsidian/transaction.py:130` sets `MAX_TRANSACTION_WRITES = 1024`, enforced at lines 3061, 3071, and 3533. Observed journal size scales at roughly 350–450 bytes per write entry, so a transaction at the sanctioned ceiling produces a journal in the region of 400 KB — about 6× the probe's limit.

In practice the probe fails at roughly 150 writes, or ~15% of the documented maximum. Measured from a real vault:

| journal bytes | writes | readable by probe |
|---|---|---|
| 124,820 | 278 | no |
| 111,548 | 275 | no |
| 50,659 | 145 | yes (barely) |
| 25,076 | 63 | yes |

Batched operations are documented usage — `skills/wiki-ingest/SKILL.md` describes ingest as "a single source or bounded batch" and lists "batch ingest" among its triggers — so ordinary use reaches this size.

Worth noting that every other bound in `hook_adapter.py` is a named module constant (`MAX_CONTEXT_BYTES`, `MAX_STATUS_BYTES`, `MAX_STATUS_ITEMS`, `MAX_TRANSACTION_SCAN`), while this one is an inline `64 * 1024` at the call site — it reads like a generic small-file bound applied to a file the package guarantees can be much larger.

## Reproduction

1. Apply any transaction with more than ~150 writes (or hand-write a journal over 64 KiB with `"state": "complete"` and a matching valid `changed-paths.json`).
2. Trigger the Stop hook, or call the probe directly:

```python
from pathlib import Path
from claude_obsidian.hook_adapter import stop_status
print(stop_status(start=Path("/path/to/vault")))
# CLAUDE_OBSIDIAN_STATUS: 1 unreadable transaction journal(s) detected.
# Run `claude-obsidian transaction recover` before the next mutation.
```

3. Confirm the journal is actually fine:

```python
import json
raw = open(journal_path, "rb").read()
print(len(raw), json.loads(raw)["state"])   # 124820 complete
json.loads(raw[:65537].decode("utf-8"))     # JSONDecodeError — what the probe sees
```

4. `claude-obsidian transaction recover --vault <vault>` returns `{"recovered": [], ...}` and the warning persists.

## Second issue: the recommended remedy cannot apply

The message directs the user to run `claude-obsidian transaction recover`. That command cannot resolve this state.

`_recover_incomplete_locked` (`transaction.py:3958`) acts only on journals whose state is in `{"prepared", "applying", "rollback-failed"}`. A journal that is already `complete` with a valid `changed-paths.json` hits the `continue` at line 3956 and is never touched. Since these journals are `complete` — and only unreadable because of the size cap — the prescribed command is a guaranteed no-op.

This is arguably the more damaging half of the bug. The warning is persistent, self-diagnosing in a way that points somewhere wrong, and the suggested fix silently succeeds while changing nothing. In our case two separate sessions concluded the vault was corrupt and invented an out-of-tree `transactions-archive/` directory to hide the oversized journals from `os.scandir` — a workaround that only worked because the scan does not recurse. Nothing prunes transaction directories, so the count only ever grows.

## Suggested fix

Any of these would resolve it:

- Raise the limit to match what transactions may legitimately produce — `MAX_TRANSACTION_RUNTIME_JSON_BYTES` (8 MiB, `transaction.py:132`) is already the package's bound for runtime JSON of this kind, and would be a consistent choice.
- Extract `state` without parsing the whole document, so journal size stops mattering to a probe that wants one key.
- Distinguish "too large to probe" from "corrupt" in the warning text, and drop the `transaction recover` suggestion for states that command cannot address.

The last point matters independently of the size cap: whatever the limit becomes, a warning should not recommend a command that provably cannot clear it.

## Environment

- claude-obsidian 2.0.0
- Python 3.14.6
- macOS (Darwin 25.5.0), arm64

STiFLeR7:
## Problem

`stop_status()` reports healthy, completed transaction journals as "unreadable" whenever the journal exceeds 64 KiB, even though they parse as valid JSON with `"state": "complete"`. The recommended remedy (`claude-obsidian transaction recover`) cannot clear the warning, since it only acts on journals whose state is `prepared`/`applying`/`rollback-failed` — a `complete` journal is skipped, so the fix silently no-ops and the count only grows.

## Root cause

`claude_obsidian/hook_adapter.py`'s journal probe read a bounded prefix through an inline `64 * 1024` cap and immediately parsed it as complete JSON:

```python
payload = _bounded_regular_bytes(root, journal, 64 * 1024)
...
state = strict_json_loads(payload.decode("utf-8")).get("state")
```

`MAX_TRANSACTION_WRITES = 1024` (`transaction.py`) permits transactions whose journals reach roughly 400 KB (~350-450 bytes/write), about 6x the probe's cap. Any journal past the cap is truncated mid-document, fails `json.loads`, and gets counted as unreadable -- in practice this triggers at ~150 writes, well inside documented usage (batch ingest, multi-file saves).

## Fix

- Read up to `MAX_TRANSACTION_RUNTIME_JSON_BYTES` (8 MiB) instead of the inline 64 KiB cap. This is already the package's own bound for runtime JSON of this shape (`transaction.py`), so it's a consistent limit rather than a new one.
- Reworded the `unreadable_journals` warning so it no longer implies `transaction recover` will fix it, since that command provably cannot touch a `complete`-state journal.

## Tests

Added two regression tests to `tests/test_hooks.py`:
- `test_stop_status_reads_large_complete_journals_within_transaction_bound` -- a 200+ KB `complete` journal (well past the old 64 KiB cap, well within the 1024-write ceiling) now reports zero warnings.
- `test_stop_status_unreadable_journal_warning_does_not_recommend_recover` -- a genuinely corrupt journal still reports as unreadable, but the message no longer points at `transaction recover`.

Both fail against the pre-fix code (verified red-before-green) and pass after.

## Verification

```
$ python3 tests/test_hooks.py
All hook tests passed.
```

Ran the full `make test` suite (test-python, test-shell, test-contracts, test-package) under WSL Ubuntu, Python 3.14.4, per `docs/windows-wsl.md`. One pre-existing failure (`test_contracts.py::test_canonical_core_capabilities_report_only_behavioral_verification`, a `wiki-lint` verifier timeout unrelated to this change) reproduces identically on unmodified `main` via `git stash` -- confirmed unrelated to this fix. Everything else passes, including `test-shell`'s 56 wiki-lock cases and `test-package`'s validator.

Closes #164

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,16 @@ implementation record for older releases.
 - The `UNSUPPORTED_PLATFORM` refusal message now points to
   `docs/windows-wsl.md` for users whose WSL setup is itself misbehaving.
 
+### Fixed
+
+- `stop_status` no longer misreports large, complete transaction journals as
+  unreadable: the probe's read cap now matches
+  `MAX_TRANSACTION_RUNTIME_JSON_BYTES` (8 MiB), the package's own bound for
+  runtime JSON of this kind, instead of an inline 64 KiB limit far below what
+  a sanctioned 1024-write transaction can produce. The "unreadable
+  transaction journal(s)" warning also no longer suggests
+  `transaction recover`, which cannot resolve that state.
+
 ## [2.1.0] - 2026-07-31
 
 Native Windows compatibility.
--- a/claude_obsidian/hook_adapter.py
+++ b/claude_obsidian/hook_adapter.py
@@ -20,6 +20,7 @@ from .paths import (
     is_relative_to,
     resolve_vault_root,
 )
+from .transaction import MAX_TRANSACTION_RUNTIME_JSON_BYTES
 
 
 MAX_CONTEXT_BYTES = 32 * 1024
@@ -273,7 +274,9 @@ def stop_status(
                 break
             journal = directory / "journal.json"
             try:
-                payload = _bounded_regular_bytes(root, journal, 64 * 1024)
+                payload = _bounded_regular_bytes(
+                    root, journal, MAX_TRANSACTION_RUNTIME_JSON_BYTES
+                )
                 if payload is None:
                     if journal.exists() or journal.is_symlink():
                         unsafe_journals += 1
@@ -296,7 +299,8 @@ def stop_status(
             warnings.append(f"{unsafe_journals} unsafe transaction journal(s) detected")
         if unreadable_journals:
             warnings.append(
-                f"{unreadable_journals} unreadable transaction journal(s) detected"
+                f"{unreadable_journals} unreadable transaction journal(s) detected "
+                "(not resolved by `transaction recover`; inspect manually)"
             )
     if not warnings:
         return ""
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -260,6 +260,37 @@ def test_stop_status_is_bounded_and_emitted_as_supported_json() -> None:
         assert payload == {"systemMessage": status}
 
 
+def test_stop_status_reads_large_complete_journals_within_transaction_bound() -> None:
+    with tempfile.TemporaryDirectory() as td:
+        base = Path(td)
+        vault = make_vault(base / "vault", "safe\n")
+        transactions = vault / ".vault-meta/transactions"
+        directory = transactions / "operation-000"
+        directory.mkdir(parents=True)
+        # A journal well past the old 64 KiB probe cap but still a valid,
+        # complete transaction record (writes legitimately scale to ~400 KB).
+        padding = "x" * (200 * 1024)
+        (directory / "journal.json").write_text(
+            json.dumps({"state": "complete", "padding": padding}),
+            encoding="utf-8",
+        )
+        status = stop_status(start=vault, environ={}, plugin_root=base / "plugin")
+        assert status == ""
+
+
+def test_stop_status_unreadable_journal_warning_does_not_recommend_recover() -> None:
+    with tempfile.TemporaryDirectory() as td:
+        base = Path(td)
+        vault = make_vault(base / "vault", "safe\n")
+        transactions = vault / ".vault-meta/transactions"
+        directory = transactions / "operation-000"
+        directory.mkdir(parents=True)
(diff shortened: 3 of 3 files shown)


STiFLeR7:
Superseded by upstream commit ad67087 ("fix: harden recovery and response boundaries"), which independently landed the same fix: raising the journal read cap from the hardcoded 64 KiB to `MAX_TRANSACTION_RUNTIME_JSON_BYTES` and no longer pointing at `transaction recover` as a blanket fix for unreadable journals. That commit's version is more thorough (also guards non-dict JSON payloads and `RecursionError`, and only recommends `transaction recover` when a journal is actually in a recognized recoverable state). Closing this PR as redundant — thanks for landing it upstream.

PR #173, opened 2026-08-25
fix: checkpoint COMMITTED_MODE_MISMATCH unsatisfiable with core.filemode=false

### LINKED ISSUE #166: checkpoint: COMMITTED_MODE_MISMATCH is unsatisfiable on any vault with core.filemode=false (Windows-created repo + WSL mutation)
## Summary

`checkpoint` can never succeed for any operation whose transaction result recorded a file mode with an executable bit, when the vault's Git repo has `core.filemode=false` — which is the default for any repo created or cloned by **native Windows Git**, even though the mutation itself ran through the supported WSL path.

The failure is permanent for that operation ID. `expected_modes` is read from the persisted transaction result, not from the filesystem, so `chmod`-ing the file afterwards cannot make the check pass.

```
ERR COMMITTED_MODE_MISMATCH: Git mode differs from the transaction result for wiki/log.md
```

## Root cause

`claude_obsidian/checkpoint.py:819` derives the expected Git mode from the stored POSIX mode:

```python
expected_git_mode = "100755" if expected_modes[relative] & 0o111 else "100644"
if entries[relative][0] != expected_git_mode:
    raise CheckpointError("COMMITTED_MODE_MISMATCH", ...)
```

and `_selected_paths` (`checkpoint.py:650`) takes that value straight from the transaction result:

```python
expected_modes[relative] = mode      # from result["modes"], persisted at apply time
```

The two halves disagree because they are produced under different rules:

| | Value | Produced by |
|---|---|---|
| Transaction result `modes` | `511` (`0o777`) | `stat` under WSL on `/mnt/c` with the `metadata` mount option, where `chmod` is real |
| Git tree entry | `100644` | Git with `core.filemode=false`, which discards the executable bit entirely |

`core.filemode=false` is not a misconfiguration. Native Windows Git sets it at `init`/`clone` because NTFS has no POSIX executable bit. A vault created on Windows and then mutated from WSL — the exact arrangement `docs/windows-wsl.md` prescribes — lands in this state by default.

## Reproduction

```bash
# Run under WSL, on a /mnt/c path mounted with `metadata`
D=/mnt/c/temp/fmrepro
rm -rf "$D"; mkdir -p "$D"; cd "$D"
git init -q .
git config core.filemode false      # what native Windows Git sets
printf 'log\n' > log.md
chmod 777 log.md                    # a transaction records mode 511 here
git add -A && git commit -q -m init

echo "core.filemode   : $(git config --get core.filemode)"   # false
echo "filesystem mode : $(stat -c '%a' log.md)"              # 777
echo "git records     : $(git ls-files -s log.md | cut -d' ' -f1)"  # 100644
echo "checkpoint wants: 100755"
```

Observed on `v2.1.0-2-g1c1bc49`, WSL Ubuntu 24.04, Python 3.12.3, vault on `/mnt/c` with `metadata,uid=1000,gid=1000`.

Control: the same script **without** `git config core.filemode false` auto-detects `filemode=true`, records `100755`, and passes. So the trigger is specifically a repo whose `core.filemode` was pinned false by native Windows Git.

## Impact

- Any vault initialised or cloned on Windows and mutated via WSL is one stray executable bit away from a permanently uncheckpointable operation.
- Because the expectation is persisted, there is no recovery path through the sanctioned tooling. The three obvious workarounds are all wrong:
  - editing `changed-paths.json` falsifies a provenance artifact so that its own verification passes;
  - `git update-index --chmod=+x` creates staged state, which `checkpoint` separately and correctly refuses;
  - committing the file `100755` records an incorrect mode for a Markdown note.
- The only remaining option is a plain `git commit`, which loses exactly what `checkpoint` exists to provide: blob-bytes verification against the transaction hashes, the lint gate, the pinned `--as-of` provenance date, and the staged-state refusal.

`--skip-lint` does not help; there is no flag that relaxes the mode assertion.

## How I hit it

A `save` transaction applied cleanly (9 paths). One pre-existing file in the vault, `wiki/log.md`, happened to be `0777` while every sibling was `0600`, so the result recorded `511` for it and `384` for the rest. The subsequent `checkpoint` refused. Five other vault files carried the same stray bit, so any later transaction touching them would have failed the same way.

The executable bits themselves look like an artifact of an earlier operation's umask rather than anything intentional — no vault file has any reason to be executable.

## Suggested fix

Any of these resolves it; (1) seems closest to the intent:

1. **Honour the repo's `core.filemode`.** When `git config --get core.filemode` is `false`, Git cannot represent the executable bit, so the mode assertion is not meaningful — compare content only. This preserves the check on platforms where it can actually hold.

2. **Compare against what Git would record**, rather than against raw `stat` output — i.e. normalise the expected mode through the same `core.filemode` rule Git applies.

3. **Record the intended mode, not the observed one.** A vault note is content; the transaction could persist `0644`/`0600` rather than whatever `stat` returned, which also stops one stray bit from poisoning a future checkpoint.

A defensive extra, independent of the above: have `transaction apply` warn when it records an executable mode for a `wiki/**` path, since that is almost always accidental and currently surfaces only much later, as an unfixable checkpoint failure.

Happy to open a PR if a preferred direction is indicated.

STiFLeR7:
## Problem

`checkpoint` can never succeed for an operation whose transaction result recorded an executable mode, when the vault's Git repo has `core.filemode=false` -- which is the default for any repo created or cloned by native Windows Git, even when the mutation itself ran through the supported WSL path.

```
ERR COMMITTED_MODE_MISMATCH: Git mode differs from the transaction result for wiki/log.md
```

The failure is permanent for that operation ID, since `expected_modes` is read from the persisted transaction result, not the filesystem, so `chmod`-ing the file afterward cannot make the check pass.

## Root cause

`claude_obsidian/checkpoint.py`'s `_verify_tree()` derives the expected Git mode from the transaction result's stored POSIX mode and compares it unconditionally against the committed tree entry:

```python
expected_git_mode = "100755" if expected_modes[relative] & 0o111 else "100644"
if entries[relative][0] != expected_git_mode:
    raise CheckpointError("COMMITTED_MODE_MISMATCH", ...)
```

`core.filemode=false` is not a misconfiguration -- native Windows Git sets it at `init`/`clone` because NTFS has no POSIX executable bit. When a vault created this way is later mutated from WSL (real `stat`), the transaction result can legitimately record an executable mode for a pre-existing file, but Git will never stage that bit and always records `100644`. The two sides can never agree.

## Fix

Skip the mode assertion when `git config --get core.filemode` reports `false`, comparing blob content only. This is not a new capability being disabled -- the check simply isn't meaningful on a repo that structurally cannot represent the bit either way. An absent or `true` value keeps the existing strict comparison.

## Tests

Added `test_executable_mode_drift_is_ignored_when_core_filemode_is_false` to `tests/test_checkpoint.py`, alongside the existing `test_executable_mode_drift_is_rejected_before_ref_update` (which continues to pass unchanged, confirming the strict path is untouched when `core.filemode` is unset).

The new test reproduces the real mechanism: `core.filemode` is disabled *before* an executable file is first staged (so the tree entry starts at `100644` and Git never records `+x`), then a `replace` transaction against that pre-existing executable file captures a real stat-based mode (`0o777`). Verified this fails with `COMMITTED_MODE_MISMATCH` against the pre-fix code and passes after the fix (red-before-green, confirmed manually against both commits since this repo's `make test-python` doesn't isolate per-test runs).

## Verification

```
$ python3 tests/test_checkpoint.py
All checkpoint tests passed.
```

Ran under WSL Ubuntu (native ext4, not a `/mnt/c` or `/mnt/d` mount -- confirmed core.filemode autodetection differs there, see note below), Python 3.14.4, per `docs/windows-wsl.md`.

Ran the full `make test-python` suite; the only failure is the pre-existing `test_contracts.py::test_canonical_core_capabilities_report_only_behavioral_verification` (`wiki-lint` verifier timeout), unrelated to this change and reproducing identically on unmodified `main`.

**Note for reviewers:** while developing this fix, I found that a repo living on a Windows-backed drvfs mount (`/mnt/c`, `/mnt/d`) under WSL2 gets `core.filemode` auto-detected as `false` by `git init` regardless of any explicit config, which made an earlier version of my test pass trivially against unfixed code. The final test and this verification run were done against a repo on native WSL ext4 (`/root/...`), where `core.filemode` autodetects `true` and the explicit `git config core.filemode false` call in the test is what actually exercises this path. Worth being aware of if CI or other contributors reproduce on a drvfs-backed WSL path.

Closes #166

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,16 @@ implementation record for older releases.
 - The `UNSUPPORTED_PLATFORM` refusal message now points to
   `docs/windows-wsl.md` for users whose WSL setup is itself misbehaving.
 
+### Fixed
+
+- `checkpoint` no longer raises an unrecoverable `COMMITTED_MODE_MISMATCH`
+  for vaults whose Git repo has `core.filemode=false` (the default for
+  repos initialized or cloned by native Windows Git, even when mutated
+  through the supported WSL path). Git cannot represent the executable bit
+  in that mode, so the committed tree entry can never match a transaction
+  result that recorded a real filesystem stat; the mode assertion is now
+  skipped when `core.filemode` is explicitly false, comparing content only.
+
 ## [2.1.0] - 2026-07-31
 
 Native Windows compatibility.
--- a/claude_obsidian/checkpoint.py
+++ b/claude_obsidian/checkpoint.py
@@ -799,6 +799,19 @@ def _tree_diff_paths(root: Path, base: str, tree: str) -> set[str]:
     return set(_decode_paths(completed.stdout))
 
 
+def _filemode_is_tracked(root: Path) -> bool:
+    """Return False only when the repo has explicitly disabled core.filemode.
+
+    Git discards the executable bit entirely when core.filemode=false (the
+    default for repos initialized or cloned by native Windows Git), so a
+    committed blob's mode can never reflect it regardless of what the
+    filesystem reports. An absent or true value keeps the strict comparison.
+    """
+
+    completed = _git(root, ["config", "--get", "core.filemode"], check=False)
+    return completed.stdout.strip() != "false"
+
+
 def _verify_tree(
     root: Path,
     *,
@@ -815,13 +828,17 @@ def _verify_tree(
             "COMMITTED_CONTENT_MISMATCH",
             "candidate tree is missing transaction paths: " + ", ".join(missing),
         )
+    check_modes = _filemode_is_tracked(root)
     for relative in paths:
-        expected_git_mode = "100755" if expected_modes[relative] & 0o111 else "100644"
-        if entries[relative][0] != expected_git_mode:
-            raise CheckpointError(
-                "COMMITTED_MODE_MISMATCH",
-                f"Git mode differs from the transaction result for {relative}",
+        if check_modes:
+            expected_git_mode = (
+                "100755" if expected_modes[relative] & 0o111 else "100644"
             )
+            if entries[relative][0] != expected_git_mode:
+                raise CheckpointError(
+                    "COMMITTED_MODE_MISMATCH",
+                    f"Git mode differs from the transaction result for {relative}",
+                )
         blob = _git_bytes(root, ["cat-file", "blob", entries[relative][1]]).stdout
         if hashlib.sha256(blob).hexdigest() != expected_hashes[relative]:
             raise CheckpointError(
--- a/tests/test_checkpoint.py
+++ b/tests/test_checkpoint.py
@@ -239,6 +239,48 @@ def test_executable_mode_drift_is_rejected_before_ref_update() -> None:
         assert note.stat().st_mode & 0o111
 
 
+def test_executable_mode_drift_is_ignored_when_core_filemode_is_false() -> None:
+    # Reproduces a vault whose Git repo was created by native Windows Git
+    # (core.filemode=false) but mutated through the supported WSL path,
+    # where `stat` on a pre-existing executable file is real. The
+    # transaction result then records the true executable mode, while Git
+    # can never stage that bit, so the two must not be compared strictly.
+    with tempfile.TemporaryDirectory() as td:
+        root = make_repo(Path(td) / "vault")
+        # core.filemode must be disabled before the executable file is first
+        # staged, matching a repo Git initialized with the bit already off:
+        # the tree entry is 100644 from the start and never records +x.
+        git(root, "config", "core.filemode", "false")
+        note = root / "wiki/Note.md"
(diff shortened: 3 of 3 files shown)


STiFLeR7:
Superseded by upstream commit ad67087 ("fix: harden recovery and response boundaries"), which independently landed the same fix: `_verify_tree` now only compares Git file-mode bits when `core.filemode` is not explicitly set to false (via a new `_filemode_is_tracked` helper), matching this PR's diagnosis of `COMMITTED_MODE_MISMATCH` being unsatisfiable with `core.filemode=false`. Closing this PR as redundant — thanks for landing it upstream.

